PR #5, opened 2025-11-10
feat: Add Cerebras LLM integration and enhance query generation

vidhanm:
- Introduced `test-query.ts` for testing natural language query generation with Cerebras.
- Updated `generate.ts` to replace 'pull' command with 'introspect' for schema retrieval.
- Enhanced CLI with `test-llm` command to test LLM connection using Cerebras or OpenAI.
- Created `test-llm.ts` to handle LLM connection testing and validation.
- Modified state machine to include optimistic and validate fields in operations.
- Improved LLM-based query generation in `llm-generator.ts` to utilize Cerebras API.
- Updated file write handler to support optimistic updates and validation functions.
- Enhanced user input handler to capture optimistic and validate fields from LLM responses.
- Added mutation generator prompt for structured mutation object generation.
- Implemented Cerebras provider in LLM module for handling LLM requests.
- Updated package dependencies to include Cerebras SDK.


demo/nextjs/components/artists-albums-view.tsx
```diff
@@ -107,6 +149,75 @@ export function ArtistsAlbumsView({
...
+        {view === "popular" && (
+          <section className="space-y-4">
+            <div className="flex items-center gap-3">
+              <h2 className="text-2xl font-semibold">Popular Artists (Popularity &gt; 80)</h2>
+              <span className="bg-purple-500 text-white px-3 py-1 rounded-full text-sm font-semibold">
+                {popularArtists.length} artists
+              </span>
+            </div>
+            <p className="text-sm text-muted-foreground">
+              🤖 Auto-generated query using Cerebras AI
```
devjain32: remove

packages/corsair/cli/watch/core/state-machine.ts
```diff
@@ -350,6 +350,8 @@ class StateMachine {
             notes: data.response.notes,
             pseudocode: (data.response as any).pseudocode,
             function_name: (data.response as any).function_name,
+            optimistic: (data.response as any).optimistic,
```
devjain32: i like the way you've thought through the `optimistic` and `validate` methods, but i'd actually recommend moving that to a new PR for now because we will want to think through that implementation. that should maybe be handled by Tanstack query

packages/corsair/cli/watch/handlers/query-generator.ts
```diff
@@ -217,14 +230,19 @@ export default ${generatedQuery.functionName};
...
+            name: operation.operationName,
+          })
+        : operationGeneratorPrompt({
+            schema,
+            type: operation.operationType,
+            name: operation.operationName,
+          })
+
+      // Get the provider from environment variable or default to "cerebras"
+      const provider = 'cerebras'
```
devjain32: we should not have a default provider but we should instead check env vars on cli watch startup. 

so, if we find a certain env var (like openai_api_key) or something like that, we should use that. if we find multiple, we should let the developer select which they want to move forward with. should not have this hardcoded here

PR #110, opened 2026-04-06
feat: add WhatsApp integration

theRizwan:
## Summary

This pull request adds a new **WhatsApp** integration package for Corsair following the repository's existing plugin conventions.

## What changed

- Added a new `@corsair-dev/whatsapp` package
- - Implemented outbound `messages.sendMessage`
- - Implemented persisted `messages.getMessages` and `messages.listConversations`
- - Added webhook handlers for inbound messages and delivery status updates
- - Added WhatsApp schema models for persisted messages and conversations
- - Wired endpoint metadata, endpoint schemas, webhook schemas, auth configuration, and webhook matching into the plugin entry point
- - Registered `whatsapp` as a provider in the shared provider constants
- - Added package-level README documentation with setup and usage examples
## Validation

- [x] `pnpm exec tsc -p packages/whatsapp/tsconfig.json --noEmit`
## Checklist

- [x] Follows existing package structure and naming conventions
- [ ] - [x] Keeps changes focused on the WhatsApp integration
- [ ] - [x] Includes setup and usage documentation

### DIFF
--- a/packages/corsair/core/constants.ts
+++ b/packages/corsair/core/constants.ts
@@ -52,6 +52,7 @@ export const BaseProviders = [
 	'twitter',
 	'twitterapiio',
 	'typeform',
+	'whatsapp',
 	'zoom',
 ] as const;
 
@@ -95,6 +96,7 @@ export type AllProviders =
 	| 'twitter'
 	| 'twitterapiio'
 	| 'typeform'
+	| 'whatsapp'
 	| 'zoom'
 	| (string & {});
 
--- a/packages/corsair/db/kysely/orm.ts
+++ b/packages/corsair/db/kysely/orm.ts
@@ -13,10 +13,10 @@ import {
 	jsonbTimestampField,
 } from './postgres';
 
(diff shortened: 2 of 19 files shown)


devjain32:
Hi, @theRizwan! Thanks for writing this plugin! A WhatsApp integration will be helpful to lots of developers.

There's a few things we found in reviewing this PR:
- query params should only be sent with `GET` requests in `client.ts`
- there's unnecessary logic in the error block in `client.ts`
- we prefer to use spread operators across api endpoints. when api's change in the future, this is one less place we have to manually update them. for example, if the whatsapp api eventually adds a new field for order_by, then we just update our zod schema. the api endpoint itself remains the same since it's just spreading what comes in from the zod schema. we sometimes manually assign fields if there's type mismatches. please take a look at any other endpoint in another plugin to see how spread operators work in api requests
- the plugin webhook matcher function should check if the origin is from whatsapp by filtering through the headers of the request (not body) unless you absolutely have to filter through body

PR #112, opened 2026-04-06
feat: add Facebook integration

theRizwan:


devjain32:
Hi, @theRizwan! Thanks for this PR! A couple of notes:
- in the `client.ts`, if the method is `GET` then only query params should be sent
- we need one more validation than `x-hub-signature-256` since github and other plugins use it. perhaps another field that's unique to facebook in the header.

Thanks again for this PR, and we'll get it merged after these changes are made!

devjain32:
@greptileai

Issue #136, opened 2026-04-09
Reddit Plugin support


NigamDevansh:
### What API or service would you like integrated?

Reddit

### API documentation link

https://www.reddit.com/dev/api/

### What would you like to do with this integration?

Full functionality to interact with reddit via agent.

### Do you need webhook support?

- [ ] Yes, I need webhook support for this integration

### Webhook details (if applicable)

Doesn't support webhooks.

PR #139, opened 2026-04-11
feat: reddit plugin

NigamDevansh:
#136


devjain32:
Hi, @NigamDevansh! Thanks for creating this PR. A few notes so far:
- the reddit `keyBuilder` function never fetches stored keys. corsair manages keys internally so it's important we retrieve a stored key if an api key is not passed in. you can take a look at any other plugin whose default auth type is `api_key` to see how to do this properly.
- i don't see where you send the auth token to the api in `client.ts`. i think the auth token is never being sent?
- i don't understand why you need `makeRequest`, `makeRedditSubredditRequest`, and the rest. can you please consolidate these down to one? if you're unable to consolidate, some comments would be helpful about why that's the case
- you don't need the export from `reddit/webhook/types.ts`
- we prefer to use spread operators across api endpoints. when api's change in the future, this is one less place we have to manually update them. for example, if the reddit api eventually adds a new field for `order_by`, then we just update our zod schema. the api endpoint itself remains the same since it's just spreading what comes in from the zod schema. we sometimes manually assign fields if there's type mismatches. please take a look at any other endpoint in another plugin to see how spread operators work in api requests
- please remove `RedditListingResponse` from all endpoint files as well. you should be using `RedditEndpointOutputSchemas['endpoint']` instead of this generic response type

Again, thank you for creating this PR! I'm excited to start using it once it's merged in!

NigamDevansh:
Hey @devjain32 , resolved the comments can you please have a look. And for the webhooks , i have take the suggestion from hacker news package , but we can remove it completely if required. And for the keybuilder , i was thinking of making it extentable for the future , when this extends to oauth version. but removed it for now and making is just for public api only.

devjain32:
Hi @NigamDevansh, that sounds good! I wanted to make some changes to the Reddit PR so I created this PR: https://github.com/NigamDevansh/corsair/pull/1

If you're good with these changes (mainly type fixes and comments), then go ahead and merge into this PR. Then we should be good to merge this into main. 

Thanks!

PR #141, opened 2026-04-11
tavily: add extract/crawl/map endpoints + api test suite

harshprasad06:
## Summary

Expands the Tavily plugin from a single `search` endpoint to four, and adds the first test suite for the package.

- **New endpoints** (all `riskLevel: 'read'` — Tavily endpoints only fetch public web data, never mutate user state):
  - `extract` — extract clean content from one or more URLs (stable)
  - `crawl` — crawl a website from a root URL with depth/breadth/category/path filters (beta)
  - `map` — map a website's URL structure (beta)
- **New test suite** — `packages/tavily/api.test.ts` mirrors the `packages/exa/api.test.ts` pattern. 17 live-API tests validate response shapes via a new `TavilyEndpointOutputSchemas` map. Closes the "zero tests" gap for tavily.
- **Bugfix** — `TavilySearchResponseSchema.answer` was typed `z.string().optional()`, but the Tavily API returns `null` when no answer is requested. Caught by the new search tests; fixed to `z.string().nullable().optional()`.

All request/response shapes use Zod and mirror the existing `search` endpoint's conventions (same client, same `logEventFromContext` pattern, same file layout under `endpoints/`).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `biome check packages/tavily` clean
- [x] `TAVILY_API_KEY=… jest` — **17/17 passing** against the live Tavily API (~19s)
  - search: 6/6 ✓
  - extract: 4/4 ✓
  - crawl: 4/4 ✓
  - map: 4/4 ✓
- [ ] Reviewer to verify beta endpoint schemas (`crawl`, `map`) are acceptable given Tavily's beta status

🤖 Generated with [Claude Code](https://claude.com/claude-code)


devjain32:
Hi, @harshprasad06! Thanks for your new PR for this plugin! I use Tavily quite a bit, so I'm excited to start using this once it's merged.

One thing missing in this PR is db interactions. As data flows through Corsair, one of the benefits of our SDK is we save information so a developer doesn't have to worry about saving it themselves. This is especially necessary for something like Tavily where you're paying per run so you likely want to cache similar runs. Please add the db schema and CRUD operations in the endpoints. You can reference other plugins to see how to do this. 

Thank you!

harshprasad06:
Hey @devjain32, thanks for the review!

Addressed the DB feedback in the latest push:

- Added a TavilyRun entity in packages/tavily/schema/database.ts (endpoint,
request, response, cachedAt) and registered it as ctx.db.runs in
schema/index.ts.

- Wired a small cache helper in packages/tavily/endpoints/cache.ts that
hashes the normalized request body (sha256 of a stable-stringified body,
prefixed by endpoint name) so key order doesn't break the hash.

- All four endpoints (search, extract, crawl, map) now do a readCachedRun
before calling the API and a writeCachedRun after a successful response, so
identical runs short-circuit before hitting the paid Tavily API.

- Followed the twitter / sharepoint plugin pattern:
ctx.db.runs.findByEntityId → upsertByEntityId, with try/catch +
console.warn around DB I/O so persistence failures never break the
endpoint. Cache hits also log cached: true via logEventFromContext so it's
observable.

  Also addressed the two Greptile comments in the same push:

- Replaced the fragile expect(typeof response.answer).toBe('string') with
expect.any(String) so a null answer fails loudly instead of silently
passing as 'object'.

- Added the missing failed_results field on TavilyCrawlResponseSchema
(optional, since the crawl endpoint is still beta).

tsc --noEmit is clean. Ready for another look 🙏

devjain32:
Hi @harshprasad06, thanks for those updates. This PR is still not quite structured how the rest of them are. I went ahead and pulled it down to refactor it myself. 

I created a PR that you can merge into this one if you're good with the changes. This PR reflects how we'd prefer to structure our code. If it looks good to you, go ahead and merge it into this PR and then we can review this one again and merge into main.

https://github.com/harshprasad06/corsair/pull/1

devjain32 (approved):

devjain32:
Hi @harshprasad06 , this PR looks good. I merged in my fixes and am going to merge this into main. Thanks again for your contribution!

PR #149, opened 2026-04-14
- Added openweathermap agent plugin into ecosystem

jaydeep-vora:
- What: New OpenWeatherMap plugin (One Call API 3.0)
- Endpoints: weather.oneCall, history.timeMachine, summary.daySummary, summary.overview
- Auth: API key via appid query parameter
- No webhooks: OpenWeatherMap is a pull-only API
- Tests: list_operations and get_schema verified in demo sandbox


devjain32:
Hi @jaydeep-vora! Thanks for writing this PR! A few things to note before we merge it in:

- we prefer to use spread operators across api endpoints. when api's change in the future, this is one less place we have to manually update them. for example, if the openweathermap api eventually adds a new field for order_by, then we just update our zod schema. the api endpoint itself remains the same since it's just spreading what comes in from the zod schema. we sometimes manually assign fields if there's type mismatches. please take a look at any other endpoint in another plugin to see how spread operators work in api requests
- please avoid type assertions (i left a comment on one as an example of what to avoid)
- please add db schema and save / update the db as data flows through. one big benefit of corsair is that we automatically save data where necessary. it's up to you what data you'd want to save, since weather information is not useful after the weather passes. however, it would be helpful to save some info (perhaps day summary?)

packages/openweathermap/endpoints/summary.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Data is available from January 2nd, 1979 to 1.5 years ahead of the current date.
+ *
+ * API: GET /onecall/day_summary
+ * Docs: https://openweathermap.org/api/one-call-3#day_summary
+ */
+export const daySummary: OpenWeatherMapEndpoints['daySummary'] = async (
+	ctx,
+	input,
+) => {
+	const query: Record<string, string | number | boolean | undefined> = {
```
devjain32: Please avoid type assertions like this one. Prefer to use spread operator here when calling the API.

devjain32:
Thanks for making these changes, @jaydeep-vora! Will be merging this soon

PR #167, opened 2026-04-21
Reduced the size of the packages...

Purujeet-git:
Massive Package Size Reduction: Reduced the size of corsair package by eleminating the integrations.By introducing Hybrid Dependency Model.
In hybrid model - Kept the Top 12 core plugins as direct dependencies and moved the rest to optional peer.
Implemented a sync re-export pattern in plugins.ts that provides helpful error messages that states that which module is the list of core plugins and which are not so the coders are not flustered.
Verified the build results.


devjain32:
Hi @Purujeet-git ! Thanks for writing this PR. This is a pretty ambitious code change. Curious why you feel this is preferred over just having a different package for each Corsair plugin?

Purujeet-git:
Because having different package for each plugin means the dev would need to import every time they need to use the different plugin since we are packaging the top 12 plugins most of the users will not need to add a new line every time keeping the code maintainable. But since the package was getting bigger in size we had to reduce the size. So if needed we can change the modules to include more which used but since not every module is used everyone but some top ones are used by many I think I went with this way.

devjain32 (approved):

PR #176, opened 2026-04-24
Feature/script tenant selector

Rudra2637:
## Summary

Adds explicit tenant selection to the Script page in Corsair Studio so users must intentionally choose the execution scope before running scripts.

Previously, the Script page relied on the app-level `tenant` state and passed it directly into `runScript`, which made tenant context implicit and could lead to running scripts against the wrong tenant.

This change aligns Script behavior with the existing Plugins page by introducing:

- Main vs Tenant scope selection
- Tenant dropdown for tenant-scoped execution
- Explicit tenant selection before running scripts
- Session-only selection (no persistence across page visits)

This improves safety for multi-tenant environments where running a script against the wrong tenant could have serious consequences.

---

### Added tenant scope controls to `ScriptPage.tsx`

Introduced:

- `scope` state (`main | tenant`)
- `activeTenant` state for selected tenant
- `tenants` state loaded from `api.listTenants()`
- `multiTenancy` check using `api.status()`

This mirrors the tenant selection pattern used in `PluginsPage.tsx`.

---

### Updated script execution behavior

Changed:

```ts
api.runScript({ code, tenant })

### DIFF
--- a/packages/studio/src/web/pages/ScriptPage.tsx
+++ b/packages/studio/src/web/pages/ScriptPage.tsx
@@ -21,7 +21,6 @@ export function ScriptPage({ tenant }: { tenant: string }) {
 	const [running, setRunning] = useState(false);
 	const [error, setError] = useState<string | null>(null);
 	const [tenants, setTenants] = useState<string[]>([]);
-	const [scope, setScope] = useState<'main' | 'tenant'>('main');
 	const [activeTenant, setActiveTenant] = useState('');
 	const [multiTenancy, setMultiTenancy] = useState(false);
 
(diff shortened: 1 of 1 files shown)


Rudra2637:
Any opinion on this pr @devjain32

devjain32:
Looks good, @Rudra2637! If multi_tenancy is enabled, I noticed the dropdown selector has `Main (integration keys)`. This technically shouldn't be possible because you should only ever be able to complete API requests as a tenant. It should always be tenant-scoped. 

We should remove the `Main (integration keys)` option from the dropdown selector. That will confuse people. What do you think?


Screenshot 2026-04-28 at 2 57 19 PM

Rudra2637:
That makes sense — I agree. Since script execution in multi-tenant mode should always be tenant-scoped, keeping the “Main (integration keys)” option is misleading and could cause confusion.

I’ve removed the main scope option and simplified the logic so users can only select a tenant before running scripts.
Have a look @devjain32

PR #240, opened 2026-06-03
feat(core): connect link for non-authenticated OAuth users

### LINKED ISSUE #237: Feature: Connect link for non-authenticated users
In `app.corsair.dev`, we have a connect link we give the user when a certain plugin isn't authenticated. For example, if they're trying to do a Gmail command and Gmail isn't authenticated, we respond with a connect link that they can go to and sign into Gmail. It allows for a graceful failure on common authentication statuses.

We need to add a similar connect link response for the open source Corsair package. If you look at the permissions structure, we have a [async message](https://github.com/corsairdev/corsair/blob/82c55a64fbaa7d5d6168e184f080a9faeab29dfe/packages/corsair/core/index.ts#L27) that we return to the agent so it's able to guide the user.

We need a similar connect link structure to give to the user. This can be returned by Corsair when we find an error contains `[auth-missing:...]`

On the actual `/connect` URL, Corsair can provide necessary oauth URLs so the user just has to wire up the UI 

Image

ambikeesshh:
Closes #237

## Summary

- Adds `AuthMissingError` class for detecting missing authentication during endpoint execution
- Adds `connect` config to `CorsairIntegration` with `baseUrl`, `redirectUri`, and optional `onAuthMissing` callback
- Intercepts auth-missing errors in endpoint binding and generates a signed connect link for the agent
- Adds `resolveConnectLink()` utility for consumer connect pages to get the full OAuth URL in one call
- Extracts OAuth state signing to `core/auth/state.ts` to avoid circular dependencies

## How it works

1. User calls `createCorsair({ plugins, kek, connect: { baseUrl, redirectUri } })`
2. When an endpoint's keyBuilder fails (empty key, missing account, missing DEK), the binding layer catches it
3. Generates a signed state and builds `${baseUrl}?plugin=${id}&state=${signedState}`
4. Returns `[auth-missing:${pluginId}] Authentication required. Direct the user to connect their account: ${connectUrl}`
5. Consumer's connect page calls `resolveConnectLink(corsair, state)` to get the full OAuth URL

## Test plan
- [x] 13 tests passing (unit + integration)
- [x] TypeScript compilation clean (zero non-test errors)
- [x] Connect link generation for empty keyBuilder, AuthMissingError, account-not-found
- [x] Custom onAuthMissing callback
- [x] Negative cases: no config, no oauthConfig, unrelated errors
- [x] OAuth state round-trip with HMAC verification

### DIFF
--- a/packages/corsair/core/endpoints/bind.ts
+++ b/packages/corsair/core/endpoints/bind.ts
@@ -48,7 +48,6 @@ function buildConnectError(
 		connectConfig.kek!,
 	);
 	const url = new URL(connectConfig.baseUrl);
-	url.searchParams.set('plugin', pluginId);
 	url.searchParams.set('state', state);
 	const connectUrl = url.toString();
 	const msg = connectConfig.onAuthMissing
--- a/packages/corsair/core/endpoints/bind.ts
+++ b/packages/corsair/core/endpoints/bind.ts
@@ -268,7 +268,12 @@ export function bindEndpointsRecursively({
 					throw err;
 				}
 
-				if (!key && connectConfig?.oauthConfig && connectConfig.kek) {
+				if (
+					!key &&
+					keyBuilder &&
+					connectConfig?.oauthConfig &&
+					connectConfig.kek
+				) {
 					const state = signState(
 						encodeOAuthState(
 							pluginId,
@@ -287,6 +292,7 @@ export function bindEndpointsRecursively({
 							: `[auth-missing:${pluginId}] Authentication required. Direct the user to connect their account: ${connectUrl}`,
 					);
 				}
+
 
 				if (!endpointHooks?.before && !endpointHooks?.after) {
 					const res = await call(0, { ...ctx, key }, args);
--- a/packages/corsair/core/endpoints/bind.ts
+++ b/packages/corsair/core/endpoints/bind.ts
@@ -242,9 +242,6 @@ export function bindEndpointsRecursively({
         let key: string | undefined
         try {
           key = keyBuilder ? await keyBuilder(ctx, 'endpoint') : undefined
-          if (!key && connectConfig?.oauthConfig && connectConfig.kek) {
-            throw new AuthMissingError(pluginId)
-          }
         } catch (err) {
           if (
             connectConfig?.oauthConfig &&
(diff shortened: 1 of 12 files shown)


ambikeesshh:
@devjain32 PTAL :)

devjain32:
@greptileai

devjain32:
Hi @ambikeesshh! Thanks for writing this PR up so quickly! I just merged it with main and ran lint so that I could review the actual core lines of code that were changed. Reviewing now!

yuvrxj-afk:
Nice work on this. Two small things I would flag here :

-  **The signed state has no expiry.** `encodeOAuthState` in `core/auth/state.ts` returns `{ plugin, tenantId }` with no `iat` or `exp`, and `verifyAndDecodeState` only verifies the HMAC. Once signed, a connect link is valid forever. If one leaks through logs, email forwarding, or browser history, it can be used weeks later to hijack the OAuth flow. Fix is small: add `iat: Date.now()` to the payload and reject in `verifyAndDecodeState` if older than a window (10–60 min). Worth setting the precedent here since #241 builds on this same state format.

-  **Connect URL malformation if `baseUrl` already has a query string.** At `bind.ts:258` and `bind.ts:284`, `` `${connectConfig.baseUrl}?plugin=...&state=...` `` produces two `?` if a dev passes `baseUrl = "https://app.example.com/connect?source=email"`. Worth using the `URL` constructor or checking for an existing `?` before appending.

Rest LGTM.

ambikeesshh:
> Nice work on this. Two small things I would flag here :
> 
> * **The signed state has no expiry.** `encodeOAuthState` in `core/auth/state.ts` returns `{ plugin, tenantId }` with no `iat` or `exp`, and `verifyAndDecodeState` only verifies the HMAC. Once signed, a connect link is valid forever. If one leaks through logs, email forwarding, or browser history, it can be used weeks later to hijack the OAuth flow. Fix is small: add `iat: Date.now()` to the payload and reject in `verifyAndDecodeState` if older than a window (10–60 min). Worth setting the precedent here since [Feature: Create Corsair Client #241](https://github.com/corsairdev/corsair/issues/241) builds on this same state format.
> * **Connect URL malformation if `baseUrl` already has a query string.** At `bind.ts:258` and `bind.ts:284`, `` `${connectConfig.baseUrl}?plugin=...&state=...` `` produces two `?` if a dev passes `baseUrl = "https://app.example.com/connect?source=email"`. Worth using the `URL` constructor or checking for an existing `?` before appending.
> 
> Rest LGTM.

working on it 
thanks for the review

packages/corsair/core/endpoints/bind.ts
```diff
@@ -23,6 +26,37 @@ export function isEndpoint(value: unknown): value is Function {
...
+): Error {
+	const state = signState(
+		encodeOAuthState(
+			pluginId,
+			connectConfig.tenantId ?? fallbackTenantId ?? 'default',
+		),
+		connectConfig.kek!,
+	);
+	const url = new URL(connectConfig.baseUrl);
+	url.searchParams.set('plugin', pluginId);
```
devjain32: We already encode the `pluginId` in the state. What's the need to add it to the URL here as well?

ambikeesshh: my bad. you are right, dropped it

devjain32: No worries! Everything else looks good to me. After merging this, I think we would need to add the logic to call the auth error in every `keyBuilder`?

ambikeesshh: yep, thats the plan. each plugin's keyBuilder would need to throw AuthMissingError when the account isn't found or the DEK is missing. the empty-key fallback covers it for now, but making it explicit per-plugin is cleaner
i can follow up with that in a separate PR

devjain32: Okay, great. This looks good to me so far. I just tested it fully. I think we're good to merge

devjain32 (approved):

PR #290, opened 2026-06-17
chore: enforce plugin validations and fix ci test configuration

Dhirenderchoudhary:
## 📝 Description

This PR strictly enforces plugin validation checks to align the repository with mature engineering standards. Specifically, it ensures that every single endpoint defined within a plugin possesses corresponding, strongly-typed risk-level metadata by verifying that `endpointMeta` correctly implements `satisfies RequiredPluginEndpointMeta`.

Additionally, this PR standardizes the test configuration for the `dodopayments` and `bitwarden` plugins to match the architecture of fully compliant plugins (e.g. Slack).

**Specific Changes:**
* Updated `scripts/validate-plugins.ts` to strictly enforce metadata and test existence.
* Added missing `webhooks.getPayloads` metadata to the `airtable` plugin.
* Fixed the `telegram` plugin so its `endpointMeta` keys perfectly match the deeply nested routing tree.
* Created standard `jest.config.cjs` and API test files for `bitwarden` and `dodopayments`.
* Fixed `PickAuth` type resolution in `bitwarden` preventing it from compiling.

## ✅ Checklist

## 🛠️ Additional Notes

This PR strictly enforces structure which could surface `Command Failed` errors during `pnpm build` or `pnpm typecheck` for future plugins if their exported endpoint groups do not completely match the documentation defined in `endpointMeta`. Future plugins MUST have these matching exactly or CI will purposely fail.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,9 @@
-## 📝 Description
+## Description
 
 <!-- Briefly describe the changes you've made and why they're necessary. -->
 <!-- Link to any related issues (e.g. "Fixes #123"). -->
 
-## ✅ Checklist
+## Checklist
 
 Before submitting your PR, please verify the following:
 
@@ -14,10 +14,10 @@ Before submitting your PR, please verify the following:
 - [ ] I have added or updated tests where applicable
 - [ ] I have added or updated necessary documentation
 
-## 📸 Screenshots / Demos (if applicable)
+## Screenshots / Demos (if applicable)
 
 <!-- If your changes affect the UI or CLI output, please include screenshots or terminal recordings. -->
 
-## 🛠️ Additional Notes
+## Additional Notes
 
 <!-- Any other information that reviewers should know? (e.g. breaking changes, new dependencies) -->
--- a/packages/airtable/index.ts
+++ b/packages/airtable/index.ts
@@ -11,6 +11,7 @@ import type {
 	PickAuth,
 	PluginAuthConfig,
 	PluginPermissionsConfig,
+	RequiredPluginEndpointMeta,
 } from 'corsair/core';
 import { AuthMissingError } from 'corsair/core';
 import { Bases, Records, Webhooks } from './endpoints';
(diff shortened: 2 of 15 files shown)


yuvrxj-afk:
Pulled this down and ran `pnpm validate:plugins` — passes on the branch, and still passes after I merged latest `main` into it (69 plugins), so none of the recent plugin merges trip the new checks. Good to confirm it won't block the queue when it lands.

Two things I'd tighten, neither blocking:

- The schema-import and `satisfies RequiredPluginEndpointMeta<` checks are raw string matches on single-quoted imports. Works today because biome normalizes quotes (grepped it — 0 plugins use double quotes), but it's matching source text instead of the module graph, so it's one formatting change away from a false fail. If it's not much effort, keying off the actual export/AST would be sturdier.
- `types.ts` is only required when `endpoints/` has >2 files. Today every plugin with an endpoints dir already has one so it never bites, but the file-count threshold means a 2-file plugin skips the check while a small 3-file one gets forced into it. I'd rather require it whenever `endpoints/` exists, or drop that check.

Bigger picture, since we're about to point the hackathon at this: can we make the failure messages consistent and link each to a plugin-authoring example? Some are one-liners, some are paragraphs — a new contributor hitting a red X wants "here's the fix / here's the reference," not a wall of text. Otherwise this is the right direction, especially dropping the tests allowlist.

Dhirenderchoudhary:
@yuvrxj-afk 
agree, I just pushed a commit to address all three of these:

- Robust matching: I swapped the raw string matching for regex (/from\s+['"]\.\/schema\/?['"]/ etc.). It now correctly handles formatting variations like double quotes or different spacing without being brittle, which is much safer than before.
- types.ts check: Dropped the weird file-count threshold entirely. Now if an endpoints/ directory exists types.ts is strictly required.
- Actionable errors: Cleaned up all the validation logs. No more walls of text every failure now throws a concise, one line error immediately followed by an exact actionable fix and points them to the slack plugin as the reference implementation.

Let me know if anything else catches your  eye

devjain32:
@greptileai

devjain32 (approved):

PR #293, opened 2026-06-17
feat: Add Webhook Tenant Matcher for all Plugins

devjain32:
## Description

Every plugin currently has a matcher that inspects the header and the body of the incoming webhook to decide if that webhook matches that plugin. However, we have no way of extracting the unique tenant id that could help us identify the tenant associated with that plugin.


devjain32:
@greptileai

devjain32:
@greptileai

PR #331, opened 2026-07-03
fix(www): disable claim button when user has active wip claim

yuvrxj-afk:
## What

After claiming an integration, the Claim buttons for all other integrations on the list page stayed enabled. Clicking one caused a Server Components render error.

## Why

`getIntegrationListForPage` overlays per-user data on top of the cached public list, but only patched `claimedByCurrentUser` — never `userCanClaim` or `wipIntegrationName`. The public caller always returns `userCanClaim: true`, so the UI had no way to know the user already had an active WIP claim.

## Fix

- `oss-public-cache.ts` — also call `getUserWipClaim` in the per-user overlay and stamp each item with the real `userCanClaim` + set `wipIntegrationName` on the list response
- `oss-sections.tsx` — pass `wipIntegrationName` down to each `IntegrationCard` so the tooltip shows correctly

### DIFF
--- a/www/src/app/oss/oss-sections.tsx
+++ b/www/src/app/oss/oss-sections.tsx
@@ -147,6 +147,7 @@ export async function OssIntegrationsSection({
 							integration={integration}
 							session={Boolean(session)}
 							index={startIndex + index + 1}
+							wipIntegrationName={integrationsData.wipIntegrationName}
 						/>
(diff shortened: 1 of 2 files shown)


yuvrxj-afk:
@greptileai

devjain32 (approved):

PR #467, opened 2026-07-16
feat(plugins): resolve managed-mode auth via Hub (Google + Outlook)

yuvrxj-afk:
## What

In managed mode, Hub holds the OAuth client credentials and issues short-lived access tokens to the app. The Google plugin auth resolvers only handled `authType: 'oauth_2'` — they tried to refresh tokens **locally**, which is impossible without the client credentials the app doesn't have. Configuring a plugin as `managed` fell through to `AuthMissingError`, so inbound managed webhooks failed to process.

## Change

Add a `managed` branch to the auth resolver of the Google OAuth plugins that obtains/refreshes the access token through the SDK's existing managed-auth helpers (`getManagedAccessToken` + `attachManagedRefreshAuth`), which refresh via Hub's `POST /oauth/refresh`. Three edits per plugin: widen the `authType` type to allow `'managed'`, import the helpers, add the branch.

Applies to **gmail, googlecalendar, googlesheets, googledrive**. The refresh machinery already existed and was exported — it was simply never called from a plugin.

## Testing

Verified end to end with Gmail managed mode: an inbound webhook that previously failed with `auth-missing:gmail:oauth_2` now processes successfully (Hub event log status `delivery_failed` → `delivered`). All four packages build clean.

## Notes

Hub requires no changes. The same 3-edit pattern generalizes to the Microsoft plugins (outlook/onedrive/sharepoint/teams) and googlemeet as a follow-up.

### DIFF
--- a/packages/gmail/index.ts
+++ b/packages/gmail/index.ts
@@ -13,6 +13,7 @@ import type {
 	RequiredPluginEndpointMeta,
 } from 'corsair/core';
 import { AuthMissingError } from 'corsair/core';
+import { attachManagedRefreshAuth, getManagedAccessToken } from 'corsair/hub';
 import { getValidAccessToken } from './client';
 import type { GmailEndpointInputs, GmailEndpointOutputs } from './endpoints';
 import {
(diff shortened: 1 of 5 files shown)


yuvrxj-afk:
Superseded by #484, which generalizes managed-mode auth into a single core mechanism — `resolveBindingKey` at both keyBuilder choke points plus a widened `PickAuth` — covering every OAuth2 plugin instead of the per-plugin resolver edits here. Closing in favor of that approach.

PR #495, opened 2026-07-23
Hub Workflow Infra

devjain32:
## Hub Workflow Infra — SDK side

The app-side execution half of workflows & triggers. The SDK receives a signed `run` envelope from Hub and executes generated workflow code against the tenant's Corsair client, streaming step results back.

### What's here
- **`workflows/execute.ts`** — `executeWorkflowRun`: runs Hub-delivered CJS in a locked-down `node:vm` realm (null-proto global, host globals absent, membrane Proxy over `corsair`). A durable `step()` / `step.sleep()` primitive gives memoized replay and resume-from-failure. Runs are time-bounded by a synchronous (vm) + wall-clock guard.
- **`tunnel/index.ts`** — the `run` tunnel type wired into delivery, gated behind `allowWorkflowExecution` (default **off** = non-breaking).
- **Backward-compat** — `TunnelAck.connectLink` restored as a deprecated alias (matches released v0.1.102 shape).
- **Tests** — signed-envelope run path + `tests/execute.test.ts` (sandbox isolation, replay + seq stability, sleep unwind/resume, both timeout guards).

### Design intent — please review against this
This is the **foundation** the rest of the workflow/trigger system builds on. We're optimizing for a **scalable architecture, not a scaled one** — no traffic numbers yet, so:
- Single-instance, gated to internal accounts.
- In-process `node:vm` isolation (documented as *not* OS-level; isolated-vm / worker / child-process is the upgrade path when we need it).
- Replay/idempotency lives durably in Postgres (Hub side); each SDK attempt is stateless.

What we want scrutinized: whether the **seams** are drawn right for scale *later* — the execution-isolation boundary, the Hub↔SDK contract, the step/replay model, and where multi-instance coordination would slot in — **not** premature optimization for load we don't have.

### Known / deferred (non-blocking)
- H1 lease-renewal overrun (>120s) — separate hardening PR.
- Per-step synchronous-loop isolation needs out-of-process execution — documented limitation of in-process `vm`.


yuvrxj-afk:
@greptile — beyond line-level correctness, could you do a **broader architectural pass** on this PR?

Context: this is the foundation the rest of the workflow/trigger system will build on, so we care most about whether the *architecture* is **scale-ready**, not whether it's already scaled. We have no load numbers yet — we're deliberately building for scalability, not premature scale. So please weight design seams over micro-optimizations.

Specifically:
- **Isolation boundary** (`workflows/execute.ts`) — is the `node:vm` sandbox seam drawn so we could swap in isolated-vm / worker / child-process later *without* changing the Hub↔SDK contract or the step model?
- **Hub↔SDK contract** (`hub/contracts/tunnel.ts`) — are the `run` envelope and `RunResultPayload` shapes stable enough to evolve? Forward-compat of new fields (e.g. `attempt`), and is versioning needed before this hardens?
- **Step / replay model** — does the memoized `step()` / `step.sleep()` design keep composing as workflows get longer or fan out, or are there dead-ends baked in?
- **Multi-instance readiness** — where would concurrency coordination slot in when we outgrow single-instance, and does anything here foreclose that path?
- **Blast radius** — on the credential + execution path, what's the worst case if a generated workflow misbehaves, and is it contained?

Not looking for premature-optimization nits — flag **architectural cul-de-sacs** and anything that'd be expensive to unwind once we're building on top of it.

yuvrxj-afk:
@greptileai — our read is that the `tenantId ?? 'default'` flag on `core/client/index.ts` is a **false positive**: single-tenant `corsair.threads.*` works end-to-end at Hub. Evidence below; happy to defer to your re-evaluation.

`'default'` is Hub's canonical single-tenant tenant id — not an unrecognized value:

1. **There's no validation to fail.** Hub's threads handler accepts `tenantId` as `z.string().min(1)` — a non-empty string, no UUID format, no tenants-table lookup (`agent-http.ts:36,41`). The only `400` is a *missing* tenantId (`:235,258`). No `SELECT … FROM tenants` exists in the threads path.
2. **It's a scope key, not a foreign key.** `agentThreads.tenantId` is `text` indexed on `(environmentId, tenantId)` (`agent.ts:24,42`); every read/write filters on that pair (`threads.ts:18`). A single-tenant app's create / list / listMessages all resolve in the same `(env, 'default')` bucket.
3. **It's Hub's own convention.** Hub normalizes to it — `connections/service.ts:153`: `input.tenantId.trim() || "default"` — and provisions `{ id: "default" }` as a real tenant. The SDK matches (`tenant-provision.ts:70`, `plugins/index.ts:400`: *"defaults to 'default' in single-tenant"*).

So `'default'` passes validation and scopes correctly end-to-end — no code change warranted here. @greptileai, please re-evaluate and adjust if you agree.

yuvrxj-afk (approved):

PR #524, opened 2026-07-30
fix(hub): restore browser delivery replay coverage for live modes

Ayush7614:
## Summary
- Rewrites the long-skipped hub browser delivery replay test off the retired `connect.status` mode onto live `auth.credentials` and `connections.sync` delivery modes
- Hardens the in-memory replay guard: validates TTL, treats expired keys as reclaimable, and documents process-local limits
- Adds dedicated unit tests for blank keys, single-use consume, TTL expiry, and whitespace trimming

## Why
`packages/corsair/tests/hub-delivery-replay.test.ts` was `it.skip`'d because `connect.status` is no longer a `BrowserDeliveryMode`. Replay protection still runs in production for every `?d=` delivery — without coverage, regressions are silent.

## Test plan
- [x] `pnpm --filter corsair test -- delivery-replay-guard.test.ts hub-delivery-replay.test.ts` (9/9 pass)
- [x] Typecheck via pre-push hook

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against replayed browser delivery tokens by rejecting previously consumed tokens.
  * Added consistent validation for missing, invalid, or blank delivery details.
  * Delivery tokens can be reused after their configured expiration period.
  * Whitespace around tokens is handled consistently, preventing accidental duplicate use.
  * Improved handling of separate delivery requests so valid tokens do not interfere with one another.

### DIFF
--- a/packages/corsair/hub/internal/delivery-replay-guard.ts
+++ b/packages/corsair/hub/internal/delivery-replay-guard.ts
@@ -46,9 +46,3 @@ export function consumeDeliveryReplayKey(
 export function resetDeliveryReplayGuardForTests(): void {
 	consumedKeys.clear();
 }
-
-/** Test helper: current in-memory key count after pruning. */
-export function countDeliveryReplayKeysForTests(): number {
-	pruneExpired(Date.now());
-	return consumedKeys.size;
-}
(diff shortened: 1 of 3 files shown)


ambikeesshh:
@greptileai review

ambikeesshh (commented):
All findings are resolved and the replay coverage looks good to me now
thanks!

devjain32 (approved):

PR #525, opened 2026-07-30
fix(www): make OSS integration claims race-safe

Ayush7614:
## Summary
- Replaces the check-then-insert OSS claim path with an atomic transaction that takes `pg_advisory_xact_lock(hashtext(integrationId))` and `FOR UPDATE` on the integration row
- Re-checks user claim eligibility inside the transaction so concurrent claimants cannot both pass the old pre-check
- Extracts a pure claim-decision helper with unit tests covering insert / idempotent re-claim / conflict cases

## Why
`integrations.claim` previously did `getLatestStatus` → if available → `insert`. Two concurrent requests could both observe an available integration and both insert `awaiting_issue`, leaving ambiguous ownership in the append-only status history.

## Test plan
- [x] `pnpm --filter @corsair/www test` (6/6 pass)
- [x] Decision matrix: null/released → insert; same user → return existing; other user → conflict

## Summary by CodeRabbit

* **New Features**
  * Added reliable integration claiming that prevents conflicting claims during simultaneous requests.
  * Existing owners can reclaim their integration without creating duplicate claim records.
  * Released integrations can be claimed again when available.

* **Bug Fixes**
  * Improved handling of active claim conflicts and ineligible claim attempts.
  * Ensured claim deadlines and status updates remain consistent.

* **Tests**
  * Added coverage for claim availability, ownership, conflicts, eligibility, and concurrent claim scenarios.

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -15,7 +15,7 @@
     "db:sync-show": "tsx src/db/sync-written-integration-show.ts",
     "db:seed-tags": "tsx src/db/seed-integration-tags.ts",
     "corsair": "tsx src/server/corsair.ts",
-    "test": "node --import tsx --test src/server/integrations/claim-integration.test.ts",
+    "test": "node --import tsx --test 'src/**/*.test.ts'",
     "sanity:schema": "sanity schema deploy",
     "sanity:deploy": "sanity deploy",
     "blog:migrate": "tsx scripts/migrate-blog-to-sanity.ts"
--- a/www/package.json
+++ b/www/package.json
@@ -15,6 +15,7 @@
     "db:sync-show": "tsx src/db/sync-written-integration-show.ts",
     "db:seed-tags": "tsx src/db/seed-integration-tags.ts",
     "corsair": "tsx src/server/corsair.ts",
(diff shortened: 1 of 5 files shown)


yuvrxj-afk:
Pushed a fix for the remaining P1: the atomic claim now takes a user-scoped advisory lock alongside the integration one (fixed user→integration order, so no deadlock), which serializes a single user claiming two different integrations at once — closing the cross-integration eligibility race. Extracted `claimLockKeys()` and added unit tests for the key derivation.

@greptileai review

yuvrxj-afk:
Addressed both points from the last review:

1. **Namespaced locks** — switched to the two-arg `pg_advisory_xact_lock(class, id)` form with distinct classes for the user vs integration locks, so the two domains share no key space and can't collide (no more relying on `hashtext` string prefixes in the single-arg space).
2. **Idempotency before eligibility** — the per-integration decision is now evaluated before the per-user eligibility gate, so a user re-claiming their own in-progress integration returns `already_owned` instead of a `PRECONDITION_FAILED` 'wip' rejection. Extracted `resolveClaimOutcome` and added unit tests for the precedence.

@coderabbitai review

yuvrxj-afk (approved):

PR #591, opened 2026-08-04
fix(slack): preserve query params for POST requests

### LINKED ISSUE #590: fix(slack): conversations.list drops cursor and limit on POST
`channels.list` calls Slack `conversations.list` as POST and puts `cursor`, `limit`, `types`, etc. in `query`.

`makeSlackRequest` strips query on anything that is not GET:

```ts
query: method === 'GET' ? query : undefined,
```

No body is sent either, so Slack gets an empty `conversations.list`. Pagination and type filters never apply.

### Fix
Forward those params (query or body, whichever Slack expects for this method) instead of dropping them.

### Check
- [ ] `cursor` / `limit` / `types` reach the outbound request
- [ ] unit test for `channels.list` request shape

```bash
pnpm --filter @corsair-dev/slack test
```

lyydsheep:
Fixes #590

## Description
Forward query parameters for non-GET Slack requests so pagination fields such as `cursor` and `limit` reach `conversations.list`. The change keeps the shared request client behavior consistent and adds a mocked regression test.

## Screenshots / Demos
This is a headless API client; the regression is exercised by the mocked Jest test and the repository CI run: https://github.com/corsairdev/corsair/actions/runs/30911390778


Dhirenderchoudhary:
LGTM

devjain32 (approved):

PR #503, opened 2026-07-24
fix(jira): sanitize trailing slashes in cloudUrl to prevent double-slash HTTP request paths 

akshatmalik-bruh:
## Description

This PR fixes a URL path formatting bug in `@corsair-dev/jira` (`packages/jira/client.ts`) where passing a `cloudUrl` containing a trailing slash (e.g., `https://mycompany.atlassian.net/`) produced double-slash request URLs (`https://mycompany.atlassian.net//rest/api/3/...`).
Fixes #502 
### Changes Made:
1. **Trailing Slash Sanitization**: Added a `sanitizeCloudUrl(url: string): string` helper inside `packages/jira/client.ts` using `url.trim().replace(/\/+$/, '')`.
2. **Helper Updates**: Applied `sanitizeCloudUrl` in `makeJiraRequest`, `uploadJiraAttachment`, and `makeJiraAgileRequest`.
3. **Automated Unit Tests**: Created `packages/jira/client.test.ts` to verify URL formatting across standard REST calls, direct attachment uploads, and Agile API calls.

Fixes double-slash HTTP 404 / 301 authorization drop errors when integrating with Jira Cloud.

## Checklist

Before submitting your PR, please verify the following:

- [x] I have run `pnpm lint` and all checks pass
- [x] I have run `pnpm typecheck` and there are no TypeScript errors
- [x] I have run `pnpm build` and all packages build successfully
- [x] I have run `pnpm test` and all tests pass
- [x] I have added or updated tests where applicable
- [x] I have added or updated necessary documentation

#### 1. Before Fix (Failing with double slashes `//`):
```text
FAIL ./client.test.ts
  Jira Client URL Sanitization
    × should not contain double slashes when cloudUrl has a trailing slash in makeJiraRequest
      Expected substring: not "atlassian.net//"
      Received string:        "https://test.atlassian.net//rest/api/3/myself"
    × should not contain double slashes when cloudUrl has a trailing slash in uploadJiraAttachment
      Expected substring: not "atlassian.net//"
      Received string:        "https://test.atlassian.net//rest/api/3/issue/PROJ-1/attachments"
    × should not contain double slashes when cloudUrl has a trailing slash in makeJiraAgileRequest
      Expected substring: not "atlassian.net//"
      Received string:        "https://test.atlassian.net//rest/agile/1.0/board"
```

#### 2. After Fix (All 3 Tests Passing Cleanly):
```text
PASS ./client.test.ts
  Jira Client URL Sanitization
[...]

PR #604, opened 2026-08-04
fix(grafana): sanitize trailing slash on baseUrl in makeGrafanaRawRequest

### LINKED ISSUE #599: fix(grafana): makeGrafanaRawRequest doesn't sanitize trailing slash on instance baseUrl
## Summary
`makeGrafanaRawRequest` builds the request URL by naive string concatenation and doesn't strip a trailing slash from the user-supplied Grafana instance URL, producing a double-slash path. This is the same bug already fixed for Jira's `cloudUrl` in #502/PR #503 — Grafana's raw-request path has the identical issue and wasn't covered by that fix (it's a separate plugin/function).

Note: `makeGrafanaRequest` (the main JSON path, in the same file) is *not* affected — it goes through the shared `corsair/async-core` request builder, which already sanitizes `BASE`. Only `makeGrafanaRawRequest`, used for the SAML/HTML endpoints, builds its URL manually.

## Current behavior
`packages/grafana/client.ts` (~line 94):
```ts
export async function makeGrafanaRawRequest(
	endpoint: string,
	bearerToken: string,
	baseUrl: string,
	options: { ... } = {},
): Promise<{ ... }> {
	const { method = 'GET', body, contentType = 'application/json' } = options;

	const url = `${baseUrl}${endpoint}`;
	...
```

Called with a leading-slash endpoint, e.g. `packages/grafana/endpoints/ring.ts:177`:
```ts
const raw = await makeGrafanaRawRequest('/ruler/ring', ctx.key, grafanaUrl);
```

If a user configures their self-hosted Grafana URL as `https://grafana.example.com/` (trailing slash — a very natural thing to paste), the resulting request goes to `https://grafana.example.com//ruler/ring`.

## Expected behavior
Sanitize `baseUrl` (strip trailing slash) before concatenating, same as the `sanitizeCloudUrl` helper added for Jira:
```ts
const cleanBaseUrl = baseUrl.replace(/\/+$/, '');
const url = `${cleanBaseUrl}${endpoint}`;
```

## Acceptance criteria
- [ ] `baseUrl` with a trailing slash produces a single-slash request path
- [ ] Existing calls without a trailing slash are unaffected
- [ ] Unit test covering the trailing-slash case (see `packages/jira/client.test.ts` for the pattern used there)

## How to verify
```bash
pnpm --filter @corsair-dev/grafana test
```

minutechreview:
## Description

`makeGrafanaRawRequest` built its request URL with naive string concatenation (`${baseUrl}${endpoint}`), so a Grafana instance URL configured with a trailing slash (e.g. `https://grafana.example.com/`) produced a double-slash request path. This fixes the same class of bug already fixed for Jira's `cloudUrl` in #503.

While fixing this, CodeRabbit's review on this PR also flagged that a non-HTTPS `baseUrl` would send the Grafana bearer token over cleartext HTTP. Since this PR was already touching the URL-building logic in that function, I fixed that too: `makeGrafanaRawRequest` now throws a `GrafanaAPIError` before calling `fetch` if the sanitized `baseUrl` isn't `https://`. Updated the function's JSDoc to document both behaviors.

I deliberately left CodeRabbit's other suggestion (an allowlist of "approved Grafana origins") out of scope — `baseUrl` is documented as pointing at a user's self-hosted Grafana instance, so there's no fixed origin to allowlist against without a larger product decision. Left a comment on the review thread explaining that.

Fixes #599

## Checklist

- [x] I have run `pnpm lint` and all checks pass (3 pre-existing warnings elsewhere in the repo, unrelated to this change and present on `main`)
- [x] I have run `pnpm build` and `@corsair-dev/grafana` builds successfully
- [x] I have run `pnpm --filter @corsair-dev/grafana test` and the unit tests I added pass (`packages/grafana/api.test.ts` is a pre-existing live-integration suite gated on real `GRAFANA_URL`/`BEARER_TOKEN` credentials that aren't available in this environment — it fails identically on `main` without them, confirmed by checking out the pre-fix file and re-running)
- [x] I have added or updated necessary documentation (JSDoc on `makeGrafanaRawRequest`)

## Screenshots / Demos

Pure request-building fix, no UI. Passing CI run for this branch (build + full test suite, all packages): https://github.com/corsairdev/corsair/actions/runs/30936881333/job/92085139297

Local test run demonstrating the fixed behavior end-to-end (mocked `fetch`, real `makeGrafanaRawRequest` call, each assertion checks the actual URL `fetch` was called with, or that it wasn't called at all for the rejected case):

```
$ pnpm --filter @corsair-dev/grafana test -- client.test.ts

PASS ./client.test.ts
  makeGrafanaRawRequest
    ✓ strips a trailing slash from baseUrl before building the request path (1 ms)
    ✓ leaves a baseUrl without a trailing slash unaffected
    ✓ refuses to send the bearer token to a non-HTTPS baseUrl (4 ms)

Tests:       3 passed, 3 total
```

## Additional Notes

No breaking changes, no new dependencies. Scope confined to `packages/grafana/client.ts` and its test file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Grafana request URL handling by removing trailing slashes before requests are sent.
  * Requests to malformed or non-HTTPS Grafana endpoints are now rejected before any network request occurs.
  * HTTPS endpoint validation now correctly accepts uppercase and mixed-case schemes.
  * Added clearer error handling when Grafana endpoint validation fails.
* **Tests**
  * Added coverage for URL normalization, endpoint validation, and prevention of invalid network requests.


minutechreview:
Thanks for the review.

- **HTTPS enforcement (quick win):** fixed in 63d5a32 — `makeGrafanaRawRequest` now throws a `GrafanaAPIError` before calling `fetch` if the sanitized `baseUrl` isn't `https://`, so the bearer token can no longer be sent in cleartext. Added a regression test.
- **Approved-origin allowlist:** leaving this out of scope for this PR. `baseUrl`/`base_url_override` is documented and expected to point at a user's self-hosted Grafana instance (see the JSDoc on `makeGrafanaRequest`), so there's no fixed "approved Grafana origin" to allowlist against here — any origin can be legitimate. Restricting that would need a product decision (e.g. an explicit per-workspace domain allowlist configured by the user), which is a bigger change than issue #599 asked for. Happy to open a follow-up issue for that if maintainers agree it's worth pursuing.

Dhirenderchoudhary:
LGTM trailing-slash strip + HTTPS reject before fetch looks right, tests cover it.

Optional : `startsWith('https://')` rejects `HTTPS://…`; `new URL(...).protocol === 'https:'` would handle that as per code rabbit reviews.

minutechreview:
Good catch — fixed in bc8a72d: now parses `baseUrl` with `URL` and compares `protocol === 'https:'` instead of the case-sensitive prefix check, with a malformed-URL case also throwing a clean `GrafanaAPIError` instead of an uncaught `TypeError`. Added regression tests for both.

Dhirenderchoudhary:
LGTM

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
@greptile review

devjain32 (approved):

PR #606, opened 2026-08-04
fix(webhooks): resolve GitHub installation_id for OAuth tenant link

yuvrxj-afk:
## Problem

GitHub returns `installation_id` in neither the OAuth token body nor the callback query — the `login/oauth/authorize` callback is just `?code=&state=`. Webhook tenant-link resolution had nothing to key on, returned null, and inbound GitHub webhooks never routed to a tenant. Slack and Linear work because their identity is in the token body.

## Change

- Propagate OAuth callback query params into tenant-link resolution via `mergeOAuthProviderData` (token body wins on collision), threaded through the tunnel payload, so query-only identity like `installation_id` can route.
- GitHub resolver falls back to `GET /user/installations` with the access token when `installation_id` is absent, using the first installation.

## Tests

- `packages/corsair/tests/oauth-callback-params.test.ts` — 3 passing
- `packages/github/webhooks/oauth-tenant-link.test.ts` — 4 passing

Pairs with corsairdev/hub `fix/webhook-tenant-link` (hub forwards the callback params and resolves `installation_id` for managed connections). Both are additive; the GitHub webhook path needs both.

## Summary by CodeRabbit

- **New Features**
  - OAuth callbacks now accept and propagate optional provider callback parameters (e.g., GitHub installation details and Slack tenant info).
  - Provider data is merged with token-derived fields taking precedence.
  - GitHub OAuth webhook tenant-linking can use token installation IDs or fall back to installation lookup.

- **Bug Fixes**
  - Tenant linking now handles missing/invalid credentials, fetch errors, and ambiguous installation results more safely.

- **Tests**
  - Added coverage for callback parameter propagation, merge precedence, and GitHub tenant-link scenarios.  

- **Chores**
  - Version bumps for related packages.


yuvrxj-afk:
@greptileai

devjain32 (approved):

Issue #594, opened 2026-08-04
fix(gitlab): webhook signature verification fails open when secret is missing


Dhirenderchoudhary:
## Summary
`verifyGitlabWebhookSignature` returns `{ valid: true }` when no secret is configured — the opposite of the fail-closed pattern used everywhere else in the codebase (Zendesk, Slack, WhatsApp, Box, etc.). Unlike Jira/Twitter (see related issues), this one skips validation unconditionally, before even looking at the `X-Gitlab-Token` header.

## Current behavior
`packages/gitlab/webhooks/types.ts` (~line 32):
```ts
export function verifyGitlabWebhookSignature(
	request: WebhookRequest<GitlabWebhookPayload>,
	secret: string,
): { valid: boolean; error?: string } {
	if (!secret) {
		return { valid: true };
	}

	const token = request.headers['x-gitlab-token'];
	...
```

## Expected behavior
```ts
if (!secret) {
	return { valid: false, error: 'Missing webhook secret' };
}
```

## Reference
Same fail-open pattern already fixed in Spotify (#519), Zoom (#520), Slack (#514), and tracked for TwitterAPI.io (#582), Monday (#581).

## Acceptance criteria
- [ ] Empty/missing secret → `{ valid: false, error: '...' }`
- [ ] Valid token still passes
- [ ] Unit test covering the empty-secret case

## How to verify
```bash
pnpm --filter @corsair-dev/gitlab test
```

## Notes
Security/consistency fix — no live GitLab credentials needed.

yuvanvk:
@Dhirenderchoudhary would like to fix this issue, could you assign it to me?

Issue #520, opened 2026-07-30
fix(zoom): return error string when webhook signing secret is missing


Dhirenderchoudhary:
## Summary
Zoom webhook verification already returns valid: false when signing secret is missing, but unlike Slack/Zendesk it omits the error field.

## Current behavior
packages/zoom/webhooks/types.ts:
if (!signingSecret) {
  return { valid: false };
}

## Expected behavior
if (!signingSecret) {
  return { valid: false, error: 'Missing webhook signing secret configuration' };
}

## Acceptance criteria
- [ ] Missing secret returns structured error message
- [ ] Optional one-line unit test

## How to verify
pnpm --filter @corsair-dev/zoom typecheck

## Notes
Same class of fix as Slack — small consistency improvement.

siddhanttiwari19:
Hi @Dhirenderchoudhary , I'd like to take this on. Confirmed in `packages/zoom/webhooks/types.ts` — the missing-secret branch returns only { valid: false }, while Slack's equivalent path already includes the structured error field, so this brings Zoom in line with that pattern.

I'll update it to return { valid: false, error: 'Missing webhook signing secret configuration' }, add the one-line unit test covering that branch, and verify with pnpm --filter @corsair-dev/zoom typecheck before pushing. Could you assign this to me?

Dhirenderchoudhary:
@siddhanttiwari19 Assigned

Issue #519, opened 2026-07-30
fix(spotify): fail closed when webhook secret is missing


Dhirenderchoudhary:
## Summary
Spotify webhook verification accepts requests when the secret is empty — opposite of fail-closed pattern used elsewhere.

## Current behavior
packages/spotify/webhooks/types.ts:
if (!secret) {
  return { valid: true };
}

## Expected behavior
if (!secret) {
  return { valid: false, error: 'Missing webhook secret' };
}

## Reference
Zendesk, Slack, Cloudinary, WhatsApp — all fail closed on missing secret.

## Acceptance criteria
- [ ] Empty/missing secret → { valid: false, error: '...' }
- [ ] Optional unit test in new or existing test file
- [ ] Consider switching require('crypto') to import from 'node:crypto' while touching the file (ESM package)

## How to verify
pnpm --filter @corsair-dev/spotify test

## Notes
Security/consistency fix — no live Spotify credentials needed.

Issue #514, opened 2026-07-30
fix(slack): return error when webhook signing secret is missing


Dhirenderchoudhary:
## Summary
verifySlackWebhookSignature rejects missing secrets but does not return an error message, unlike other webhook plugins.

## Current behavior
In packages/slack/webhooks/types.ts, when !signingSecret:
return { valid: false };

Webhook handlers fall back to a generic "Signature verification failed" message.

## Expected behavior
Return a clear error string, consistent with Zendesk/Box/etc:
return { valid: false, error: 'Missing webhook secret' };

## Reference
- packages/zendesk/webhooks/types.ts — missing secret returns { valid: false, error: '...' }
- packages/slack/webhooks/messages.ts — uses verification.error || 'Signature verification failed'

## Acceptance criteria
- [ ] Missing/empty signing secret returns { valid: false, error: '...' }
- [ ] Optional: add packages/slack/webhooks.test.ts with one missing-secret case

## How to verify
pnpm --filter @corsair-dev/slack test

## Notes
No live Slack credentials needed for the fix or a basic unit test.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

PR #608, opened 2026-08-05
fix(jira): fail closed when webhook secret or signature is missing

### LINKED ISSUE #595: fix(jira): webhook signature verification fails open when secret and signature header are both missing
## Summary
`verifyJiraWebhookSignature` accepts a request with no `x-hub-signature` header as long as no secret is configured — an attacker just needs to omit the signature header to bypass verification entirely.

## Current behavior
`packages/jira/webhooks/types.ts` (~line 183):
```ts
if (!signatureHeader) {
	// Only allow unsigned requests when no secret has been configured
	if (!secret) {
		return { valid: true };
	}
	return { valid: false, error: 'Missing x-hub-signature header' };
}
```

## Expected behavior
```ts
if (!signatureHeader) {
	return { valid: false, error: 'Missing x-hub-signature header' };
}
if (!secret) {
}
```
(Order doesn't matter much — the point is neither missing-secret nor missing-header should ever resolve to `valid: true`.)

## Reference
Same fail-open family as #519 (Spotify), #520 (Zoom), #514 (Slack).

## Acceptance criteria
- [ ] Missing signature header → `valid: false` regardless of secret configuration
- [ ] Missing secret → `valid: false`
- [ ] Valid signature with configured secret still passes
- [ ] Unit test covering both missing-header and missing-secret cases

## How to verify
```bash
pnpm --filter @corsair-dev/jira test
```

## Notes
Security/consistency fix — no live Jira credentials needed.

sushantlokhande14:
## Description

Fixes #595.

`verifyJiraWebhookSignature` returned `{ valid: true }` for any request with no `x-hub-signature` header, as long as no secret was configured — so omitting the signature header bypassed verification entirely.

```ts
// before
if (!signatureHeader) {
	// Only allow unsigned requests when no secret has been configured
	if (!secret) {
		return { valid: true };   // <-- fail open
	}
	return { valid: false, error: 'Missing x-hub-signature header' };
}
```

Now the secret is checked first and a missing header is never valid, so neither a missing secret nor a missing signature can resolve to `valid: true`:

```ts
// after
if (!secret) {
}
...
if (!signatureHeader) {
	return { valid: false, error: 'Missing x-hub-signature header' };
}
```

I deliberately matched the ordering and error strings already used by `verifySpotifyWebhookSignature`, since that verifier was fixed for this same fail-open family (#519, #520, #514) — so all the plugins now read the same way.

Adds `packages/jira/webhooks/types.test.ts`, mirroring the Spotify verifier's test layout:

| case | expected |
|---|---|
| secret missing (header present) | `Missing webhook secret` |
| **secret and header both missing** | `valid: false` — this is the actual bypass |
| header missing (secret present) | `Missing x-hub-signature header` |
| signature signed with a different key | `Invalid signature` |
| signature of the wrong length | `Signature length mismatch` |
| correct signature round-trip | `valid: true` |
| header supplied as `string[]` | `valid: true` |

All four acceptance criteria in the issue are covered.

## Checklist

- [x] I have run `pnpm lint` and all checks pass — `biome check` is clean on both changed files, and the repo's own lint-staged hook (`biome check --write`) ran clean on both commits
- [x] I have run `pnpm typecheck` and there are no TypeScript errors — `tsc --build` exits 0; the repo's pre-push typecheck hook also passed
- [x] I have run `pnpm build` and all packages build successfully — `pnpm --filter @corsair-dev/jira build` succeeds
- [x] I have run `pnpm test` and all tests pass — green in CI; **see note below** for local runs

Two notes so the ticks above are not taken on trust:

**Tests.** My new suite passes 7/7, but the pre-existing `packages/jira/api.test.ts` fails 12/12 on a clean checkout of `main` too, because it builds an auth header from an unset API key (`Buffer.from(apiKey)` → `TypeError: ... Received undefined`). Unrelated to this change; I left it alone. Happy to fix it in a separate PR if useful.

**Lint.** I verified `biome check` on the two files I changed rather than repo-wide, because I'm on a Windows checkout where `core.autocrlf` gives every file CRLF and repo-wide `biome check .` therefore fails identically on unmodified `main`. Git normalises to LF on commit, so the diff itself carries no line-ending changes — `git diff --stat` on the branch is `+117 / -4` across exactly two files.

```
# on clean main:            Test Suites: 1 failed, 1 total   Tests: 12 failed, 12 total
# on this branch:           Test Suites: 1 failed, 1 passed  Tests: 12 failed, 7 passed
#                                                     ^ pre-existing        ^ new, all passing
```

On R4 in `PLUGIN_PR_RULES.md`: the rule exists because "neither CI nor review bots can call the real third-party API." That doesn't apply here — this changes a pure local HMAC comparison, calls no Jira endpoint, and needs no credentials (the issue notes this too). The unit tests are the verification, and they run in CI. Terminal output:

```
$ npx jest --config packages/jira/jest.config.cjs webhooks/types.test.ts

 PASS  packages/jira/webhooks/types.test.ts
  verifyJiraWebhookSignature
    ✓ should fail closed when secret is missing
    ✓ should fail closed when both secret and signature header are missing
    ✓ should return invalid if signature header is missing
    ✓ should return invalid if signature does not match
    ✓ should return invalid if the signature length does not match
    ✓ should return valid for a correct signature round-trip
    ✓ should accept the signature header when provided as an array

```

**Working proof:** the CI run for this branch, where the suite executes and passes — [`CI Checks`](https://github.com/corsairdev/corsair/actions/runs/30980617048/job/92224037009).

To be straight about it: that is a link to a CI job, not a screen recording. I can't record one, and for this change there is nothing to film — no UI, no CLI output, no third-party call. If a recording is required regardless, say so and I'll arrange it.

## Additional Notes

- Scope is `packages/jira/**` only (R1) — no changes to `constants.ts` or the lockfile were needed.
- No breaking change for correctly-configured webhooks: a request with a valid signature and a configured secret behaves exactly as before. The behaviour that changes is precisely the insecure one — a deployment relying on unsigned webhooks with no secret set will now be rejected, which is the point of the fix.
- #594 is the same fail-open bug in the GitLab verifier; I'll send that as a separate PR to keep one plugin per PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jira webhook signature validation to fail fast when the webhook secret is missing.
  * Added clearer validation handling for missing, invalid, mismatched, and array-based signature headers.
  * Verified successful validation for correctly signed webhook requests.

### DIFF
--- a/packages/jira/webhooks/types.test.ts
+++ b/packages/jira/webhooks/types.test.ts
@@ -14,15 +14,19 @@ describe('verifyJiraWebhookSignature', () => {
 	const sign = (key: string): string =>
 		`sha256=${crypto.createHmac('sha256', key).update(rawBody).digest('hex')}`;
 
+	const makeRequest = (
+		headers: Record<string, string | string[] | undefined>,
+	): WebhookRequest<JiraWebhookPayload> => ({
+		payload,
+		headers,
+		rawBody,
+	});
+
 	it('should fail closed when secret is missing', () => {
-		const request: WebhookRequest<unknown> = {
-			payload,
-			headers: {
-				'x-hub-signature': sign(secret),
-			},
-			rawBody,
-		};
-		const result = verifyJiraWebhookSignature(request, '');
+		const result = verifyJiraWebhookSignature(
+			makeRequest({ 'x-hub-signature': sign(secret) }),
+			'',
+		);
 		expect(result).toEqual({
 			valid: false,
 			error: 'Missing webhook secret',
@@ -33,22 +37,15 @@ describe('verifyJiraWebhookSignature', () => {
 		// The regression: omitting the signature header used to return
 		// { valid: true } whenever no secret was configured, so an attacker
 		// could bypass verification entirely by simply not sending a signature.
-		const request: WebhookRequest<unknown> = {
-			payload,
-			headers: {},
-			rawBody,
-		};
-		const result = verifyJiraWebhookSignature(request, '');
-		expect(result.valid).toBe(false);
+		const result = verifyJiraWebhookSignature(makeRequest({}), '');
+		expect(result).toEqual({
+			valid: false,
+			error: 'Missing webhook secret',
+		});
 	});
 
 	it('should return invalid if signature header is missing', () => {
-		const request: WebhookRequest<unknown> = {
-			payload,
-			headers: {},
-			rawBody,
-		};
-		const result = verifyJiraWebhookSignature(request, secret);
+		const result = verifyJiraWebhookSignature(makeRequest({}), secret);
 		expect(result).toEqual({
 			valid: false,
 			error: 'Missing x-hub-signature header',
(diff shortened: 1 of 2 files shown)


Dhirenderchoudhary:
@greptile review

devjain32 (approved):

PR #610, opened 2026-08-05
fix(monday): verify webhook signatures and fail closed on missing secret

### LINKED ISSUE #581: fix(monday): webhook handlers skip signature verification
Monday webhook handlers never verify signatures. The check is commented out in `item-created.ts`, `status-changed.ts`, and `column-value-changed.ts`, then the handler keeps going and can write to the DB.

`verifyMondayWebhookSignature` already exists in `webhooks/types.ts`. Just not called.

Even with a secret configured, forged events get through.

### Fix
Call the verifier in all three handlers before any DB write. Return 401 on failure.

While you're in `types.ts`, empty secret currently returns `{ valid: true }`. Flip that to fail closed like the other plugins.

### Check
- [ ] all three handlers verify before writing
- [ ] bad/missing signature → 401, no DB write
- [ ] empty secret fails closed
- [ ] small unit test for the reject path

```bash
pnpm --filter @corsair-dev/monday test
```

sushantlokhande14:
## Description

Fixes #581.

Two related problems, as described in the issue:

**1. The handlers never verified anything.** In `item-created.ts`, `status-changed.ts` and `column-value-changed.ts` the verification block was present but commented out, and execution fell straight through to the event logic — so any request matching the event shape was accepted, and `itemCreated` went on to write to the database.

```ts
// before — in all three handlers
// const verification = verifyMondayWebhookSignature(request, ctx.key);
// if (!verification.valid) { ... 401 ... }

const event = request.payload.event;   // <-- reached unconditionally
```

The block is now enabled in each handler, returning 401 before any DB write.

**2. `verifyMondayWebhookSignature` failed open on an empty secret.** `if (!secret) return { valid: true }` → now returns `Missing webhook secret`, matching the string and shape already used by the Spotify verifier after the same family was fixed there (#519, #520, #514).

### `challenge.ts` is deliberately left alone

It has its own matcher (`createMondayChallengeMatch`) and is the subscription handshake Monday sends when the webhook is first registered — before any secret is exchanged — so verifying it would make webhook setup impossible. It performs no DB write and only echoes the challenge token back. Flagging it explicitly since "all handlers" could be read to include it.

## Tests

Adds `packages/monday/webhooks/webhooks.test.ts` (9 tests, all passing):

**Verifier** — missing secret → `Missing webhook secret`; missing `Authorization` header; missing raw body; correctly-signed request → valid.

**Handlers** — for each of the three: a signature computed with a different key returns `success: false` / `statusCode: 401`. For `itemCreated` specifically:

- forged signature → 401 **and** `ctx.db.items.upsertByEntityId` is never called
- no secret configured → rejected **and** no upsert
- correctly signed → `success: true` and the item is upserted with the expected payload

That covers all four boxes in the issue, including "no DB write" on the reject path.

`packages/monday/jest.config.cjs` maps `corsair/http` but not `corsair/core`. The handlers import `logEventFromContext` from `corsair/core`, so without that mapping the suite cannot import a handler at all — it dies with `Cannot find module 'corsair/core'` before running a single assertion. I added the one mapping, copying it verbatim from `packages/cloudinary/jest.config.cjs`, which is the plugin that already does handler-level webhook tests. Still inside `packages/monday/**`, so R1 scope holds.

## Checklist

- [x] I have run `pnpm lint` and all checks pass — `biome check` clean on changed files; the repo's lint-staged hook ran clean on all three commits
- [x] I have run `pnpm typecheck` and there are no TypeScript errors — `tsc --build packages/monday` exits 0

**On the test box.** My suite passes 9/9. `packages/monday/api.test.ts` fails 11/11 on a clean checkout of `main` as well (it builds an auth header from an unset API key), so this branch changes nothing there:

```
clean main:   Test Suites: 1 failed, 1 total            Tests: 11 failed, 11 total
this branch:  Test Suites: 1 failed, 1 passed, 2 total  Tests: 11 failed, 9 passed, 20 total
                                     ^ +1 (mine)               ^ identical  ^ +9 (mine)
```

No UI. Local JWT verification only. Evidence is the unit tests + issue:
https://github.com/corsairdev/corsair/issues/581

## Additional Notes

- Scope is `packages/monday/**` only (R1).
- **This one is behaviour-changing in a way the other fail-open fixes were not**, so worth calling out: any existing Monday deployment that has webhooks running *without* a configured secret will start getting 401s. That is the intent of the issue, but it is a live-traffic change rather than a purely theoretical hardening — you may want it noted in release notes.
- Companion PRs in the same family: #608 (Jira, #595) and #609 (GitLab, #594). Kept separate for one-plugin-per-PR.

## Summary by CodeRabbit

* **Security**
  * Monday webhook authorization signatures are now verified using HS256 JWT validation.
  * Invalid, expired, malformed, or unauthorized requests receive a 401 response.
  * Bearer authorization is accepted regardless of capitalization.

* **Bug Fixes**
  * Prevented unverified webhook events from being processed or persisted.

* **Tests**
  * Added coverage for valid tokens, forged signatures, expired tokens, missing secrets, Bearer authorization, and rejected requests.

### DIFF
--- a/packages/monday/jest.config.cjs
+++ b/packages/monday/jest.config.cjs
@@ -44,6 +44,7 @@ module.exports = {
 		],
 	},
 	moduleNameMapper: {
+		'^corsair/core$': '<rootDir>/../corsair/core.ts',
 		'^corsair/http$': '<rootDir>/../corsair/http.ts',
 		'^(\\.\\.?/.*)\\.js$': '$1',
 	},
--- a/packages/monday/webhooks/column-value-changed.ts
+++ b/packages/monday/webhooks/column-value-changed.ts
@@ -1,19 +1,19 @@
 import { logEventFromContext } from 'corsair/core';
 import type { MondayWebhooks } from '../index';
-import { createMondayMatch } from './types';
+import { createMondayMatch, verifyMondayWebhookSignature } from './types';
 
 export const columnValueChanged: MondayWebhooks['columnValueChanged'] = {
 	match: createMondayMatch('change_column_value'),
 
 	handler: async (ctx, request) => {
-		// const verification = verifyMondayWebhookSignature(request, ctx.key);
-		// if (!verification.valid) {
-		// 	return {
-		// 		success: false,
-		// 		statusCode: 401,
-		// 		error: verification.error || 'Signature verification failed',
-		// 	};
-		// }
+		const verification = verifyMondayWebhookSignature(request, ctx.key);
+		if (!verification.valid) {
+			return {
+				success: false,
+				statusCode: 401,
+				error: verification.error || 'Signature verification failed',
+			};
+		}
 
 		const event = request.payload.event;
 
--- a/packages/monday/webhooks/item-created.ts
+++ b/packages/monday/webhooks/item-created.ts
@@ -1,19 +1,19 @@
 import { logEventFromContext } from 'corsair/core';
 import type { MondayWebhooks } from '../index';
-import { createMondayMatch } from './types';
+import { createMondayMatch, verifyMondayWebhookSignature } from './types';
 
 export const itemCreated: MondayWebhooks['itemCreated'] = {
 	match: createMondayMatch('create_pulse'),
 
 	handler: async (ctx, request) => {
-		// const verification = verifyMondayWebhookSignature(request, ctx.key);
-		// if (!verification.valid) {
-		// 	return {
-		// 		success: false,
-		// 		statusCode: 401,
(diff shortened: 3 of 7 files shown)


ambikeesshh:
@greptileai review

ambikeesshh (commented):
verifier was HMAC against body, monday board webhooks are HS256 JWT
fixed that + bearer casing + fail-closed

lgtm now. you can merge it @devjain32

devjain32 (approved):

PR #626, opened 2026-08-09
fix(outlook): reject an empty value array during clientState verification

### LINKED ISSUE #625: fix(outlook,teams): clientState verification fails open on an empty value[] array
## Summary

`verifyOutlookWebhookSignature` and `verifyTeamsClientState` both validate the clientState of every notification with `Array.prototype.every`. `[].every(...)` is `true`, so a request whose `value` array is **empty** passes verification without carrying any clientState at all.

Found while working on #585 (the SharePoint half of this family); filing separately since it's a different plugin and a different root cause from the missing-secret variants.

## Current behavior

`packages/outlook/webhooks/types.ts` (~line 130):
```ts
const notifications = request.payload?.value ?? [];
);

if (!allMatch) {
	return { valid: false, error: 'Client state mismatch' };
}

return { valid: true };   // <-- reached for value: []
```

`packages/teams/webhooks/types.ts` (~line 198) has the same shape:
```ts
);
```

So `{ "value": [] }` — or, for outlook, a body with no `value` key at all — returns `{ valid: true }` from a function whose entire job is to reject unverified requests.

`packages/onedrive` is **not** affected: it checks a single notification and rejects a missing clientState explicitly.

## Impact — deliberately not overstating this

Neither plugin writes to the database off an empty notification list, so this is not data exfiltration:

- **outlook**: `verifyAndExtract` finds no `value[0]` and returns `{ success: true, data: undefined }` — so an unauthenticated request gets a 200 instead of a 401.
- **teams**: `channel.ts` proceeds past verification and calls `await ctx.keys.get_access_token()` before iterating the (empty) list — so an unauthenticated request causes a token fetch/refresh, then a 200.

The reason it's still worth fixing is that the verifier is the security boundary, and it currently answers "yes, verified" to a request that carried no credential. Any future code added between the verification and the length check inherits that. It is also inconsistent with the fail-closed direction the rest of the plugins have been moving in (#519, #520, #514, #594, #595, #581, #585).

## Expected behavior

An empty or missing `value` array should be rejected, as SharePoint now does after #585:

```ts
const notifications = request.payload?.value;
if (!Array.isArray(notifications) || notifications.length === 0) {
	return { valid: false, error: 'Invalid payload: missing value array' };
}
```

## Note on the validation handshake

Worth stating explicitly, because it is the trap in this area: Microsoft Graph's subscription-validation request arrives with no notifications, so a naive fix could break webhook setup.

It does not here — outlook answers the handshake in a **separate webhook** (`packages/outlook/webhooks/validation.ts`) that never reaches these verifiers, and teams matches on resource patterns that an empty body cannot satisfy. So rejecting an empty `value[]` is safe for both. (SharePoint was different: it answers the handshake *inside* the notification handler, which is why #585 needed care there.)

## Acceptance criteria

- [ ] `{ value: [] }` → `valid: false` in both verifiers
- [ ] outlook: a body with no `value` key → `valid: false`
- [ ] a populated, correctly-matching `value[]` still passes
- [ ] the Graph validation handshake still completes
- [ ] unit tests for the empty-array path in both plugins

## How to verify

```bash
pnpm --filter @corsair-dev/outlook test
pnpm --filter @corsair-dev/teams test
```

Security/consistency fix — no live Microsoft credentials needed. Happy to send the two PRs (one per plugin, per R1) if you'd like this fixed.

sushantlokhande14:
## Description

Fixes #625 (outlook half — the teams half is a separate PR, per R1).

`verifyOutlookWebhookSignature` validates every notification's clientState with `Array.prototype.every`, and `[].every(...)` is `true`. So a request whose `value` array was empty — or which had no `value` key at all — carried no clientState and was still reported as verified.

```ts
// before
const notifications = request.payload?.value ?? [];
);
if (!allMatch) {
	return { valid: false, error: 'Client state mismatch' };
}
return { valid: true };   // <-- reached for value: []
```

```ts
// after
const notifications = request.payload?.value;
if (!Array.isArray(notifications) || notifications.length === 0) {
	return { valid: false, error: 'Invalid payload: missing value array' };
}
```

Same guard and error string the SharePoint verifier uses, so the Graph plugins read the same way.

I found this while working on #585 and want to be straight about the severity rather than dress it up: **this is hardening, not an exploitable data path.** `verifyAndExtract` finds no `value[0]` and returns `{ success: true, data: undefined }`, so nothing is read or written — the concrete effect is that an unauthenticated request gets a 200 instead of a 401. It's worth fixing because the verifier is the security boundary and currently answers "verified" to a request that carried no credential, which any code added between the check and the `value[0]` lookup would inherit.

## The handshake, explicitly

Rejecting an empty `value[]` is the kind of change that can break Microsoft Graph's subscription-validation request, which arrives with no notifications. It doesn't here: outlook answers that handshake in a **separate webhook**, `packages/outlook/webhooks/validation.ts`, which never reaches this verifier.

(SharePoint was the opposite — it answers the handshake *inside* the notification handler, which is why #585 needed care there. Flagging the difference since these two look interchangeable.)

## Tests

`packages/outlook/webhooks/types.test.ts`, 7 tests, all passing:

| case | expected |
|---|---|
| **`value: []`** | `Invalid payload: missing value array` — the bug |
| **no `value` key** | `valid: false` |
| missing clientState argument | `Missing client state` |
| clientState mismatch | `Client state mismatch` |
| batch where a later notification is forged | `Client state mismatch` |
| notification with no clientState of its own | `Client state mismatch` |
| every notification matches | `valid: true` |

## Checklist

- [x] I have run `pnpm typecheck` and there are no TypeScript errors — `tsc --build packages/outlook` clean

**On the test box.** My suite passes 7/7. `packages/outlook/api.test.ts` fails 22/22 on clean `main` too (it needs live credentials), unchanged here:

```
clean main:   Test Suites: 1 failed, 1 passed, 2 total   Tests: 22 failed,  4 passed, 26 total
this branch:  Test Suites: 1 failed, 2 passed, 3 total   Tests: 22 failed, 11 passed, 33 total
                                     ^ +1 (mine)                ^ identical  ^ +7 (mine)
```

Per R4's rationale ("neither CI nor review bots can call the real third-party API") — this is a local array check, calls no Graph endpoint, and needs no credentials. The tests are the verification and they run in CI:

```
$ pnpm --filter @corsair-dev/outlook test

 PASS  webhooks/types.test.ts
  verifyOutlookWebhookSignature
    √ should fail closed when the client state is missing
    √ should reject an empty value array
    √ should reject a payload with no value key
    √ should reject when the client state does not match
    √ should reject a batch where a later notification is forged
    √ should reject a notification with no client state of its own
    √ should accept when every notification carries the expected client state
```

**Working proof:** the CI run for this branch is the same evidence — it executes this suite on your infrastructure rather than my machine. Repo CI: https://github.com/corsairdev/corsair/actions

## Additional Notes

- Scope is `packages/outlook/**` only (R1).
- Not a breaking change for real traffic: a genuine Graph notification always carries at least one entry in `value`. What stops being accepted is a request with none, which was never a legitimate notification.
- `packages/onedrive` is **not** affected — it checks a single notification and rejects a missing clientState explicitly. Only outlook and teams share this shape.
- Companion PR for the teams half of #625 follows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Outlook webhook validation by rejecting missing, empty, malformed, or invalid notification payloads with clear errors.
  * Strengthened signature verification against missing, mismatched, or forged client states.
  * Improved notification matching for null entries, invalid JSON, and newly created messages.
  * Invalid contact, event, and message webhook payloads now return an appropriate authorization failure instead of incomplete results.

* **Tests**
  * Added comprehensive coverage for invalid payloads, forged notifications, null entries, malformed data, and successful matching scenarios.


packages/outlook/webhooks/types.ts
```diff
@@ -127,7 +127,14 @@ export function verifyOutlookWebhookSignature(
 		return { valid: false, error: 'Missing client state' };
 	}
 
-	const notifications = request.payload?.value ?? [];
+	// An empty array must not pass: [].every(...) is true, so a request that
+	// carried no notification -- and therefore no clientState -- would otherwise
+	// be reported as verified.
+	const notifications = request.payload?.value;
+	if (!Array.isArray(notifications) || notifications.length === 0) {
```
ambikeesshh, line 134: this still throws on `value: [null]`. please reject non-object entries before reading `clientState`, and add the same guard in `createOutlookMatch` since it reads `notification.resource` first.

packages/outlook/webhooks/types.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		...overrides,
+	} as OutlookChangeNotification;
+}
+
+function makeRequest(payload: unknown): WebhookRequest<OutlookWebhookPayload> {
+	return {
+		payload,
+		rawBody: JSON.stringify(payload),
+		headers: {},
+	} as unknown as WebhookRequest<OutlookWebhookPayload>;
```
ambikeesshh, line 24: can we add a short comment explaining why this `unknown` cast is needed? this is the typing gap greptile flagged.

packages/outlook/webhooks/types.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		);
+		expect(result).toEqual({
+			valid: false,
+			error: 'Invalid payload: missing value array',
+		});
+	});
+
+	it('should reject a payload with no value key', () => {
+		const result = verifyOutlookWebhookSignature(makeRequest({}), CLIENT_STATE);
+		expect(result.valid).toBe(false);
```
ambikeesshh, line 52: can we assert the full error result here too? that keeps missing and empty payloads on the same contract.

ambikeesshh (commented):
the empty-array hardening looks good, but malformed entries still throw in both the matcher and verifier. please address the inline notes and add the R4 recording, then this should be ready for another pass.

ambikeesshh:
@greptileai

ambikeesshh:
@greptileai review

ambikeesshh:
image

ambikeesshh (commented):
lgtm

devjain32 (approved):

PR #629, opened 2026-08-09
fix(telegram): fail closed when no webhook secret token is configured

### LINKED ISSUE #628: fix(telegram): webhook verification fails open when no secret token is configured, allowing forged messages to be persisted
## Summary

## Current behavior

```ts
	secretToken: string,
	if (!secretToken) {
		return { valid: true };
	}
	...
```

```ts
	{
		...event.message,
	},
);
```

## Expected behavior

```ts
if (!secretToken) {
}
```

## Acceptance criteria

## How to verify

```bash
```

## Note on the comparison

sushantlokhande14:
## Description

Fixes #628.

`verifyTelegramWebhookSignature` returned `{ valid: true }` whenever no secret token was configured, so any caller who knew the webhook URL was treated as Telegram.

```ts
// before
if (!secretToken) {
	return { valid: true };
}
```
```ts
// after
if (!secretToken) {
}
```

One line, matching the shape and `'Missing webhook secret'` string adopted across this family (#594, #595, #581, #585).

### Why this one has teeth

I found this by grepping the remaining plugins for the same shape after the earlier fixes, and it is the one instance where the fail-open leads somewhere. The other plugins in this family stopped short of acting on the unverified request. `packages/telegram/webhooks/message.ts` does not:

```ts
	{
		...event.message,
	},
);
```

Every value there comes from the request body. Against a deployment with no secret token, an unauthenticated caller can insert rows with an arbitrary `message_id`, `chat_id`, `authorId`, text and timestamp — forging messages attributed to a chosen user in a chosen chat. Because it's an upsert keyed on `message_id`, it can also **overwrite a genuine message**. The other nine handlers call the same verifier first.

Telegram's `secret_token` is an *optional* `setWebhook` parameter, so an unset secret is the default rather than an exotic misconfiguration.

## Tests

`packages/telegram/webhooks/webhooks.test.ts`, 8 tests, all passing.

**Verifier** — missing secret → `Missing webhook secret`; missing header; mismatched token; matching token; array-valued header.

**Handler** — and this is the part that matters:

- no secret configured → 401 **and** `ctx.db.messages.upsertByEntityId` never called
- forged token → 401 **and** no upsert
- correct token → success, message persisted with the expected id and chat_id

## Checklist

- [x] I have run `pnpm typecheck` and there are no TypeScript errors — `tsc --build packages/telegram` clean

**On the test box.** My suite passes 8/8. The package's two pre-existing suites fail identically on clean `main` (`api.test.ts` needs live credentials, `integration.test.ts` needs a workspace build), unchanged here:

```
clean main:   Test Suites: 2 failed, 2 total            Tests: 12 failed, 1 passed, 13 total
this branch:  Test Suites: 2 failed, 1 passed, 3 total  Tests: 12 failed, 9 passed, 21 total
                                     ^ +1 (mine)               ^ identical  ^ +8 (mine)
```

`packages/telegram/jest.config.cjs` mapped `corsair/http` but not `corsair/core`. The handler imports `logEventFromContext` from `corsair/core`, so without that mapping the suite can't import a handler at all — it dies with `Cannot find module 'corsair/core'` before a single assertion. Added the one mapping, copied from `packages/cloudinary/jest.config.cjs`, which already does handler-level webhook tests. Still inside `packages/telegram/**`, so R1 holds.

Per R4's rationale ("neither CI nor review bots can call the real third-party API") — this is a local string comparison, calls no Telegram endpoint, and needs no credentials. The tests are the verification and they run in CI:

```
$ pnpm --filter @corsair-dev/telegram test

 PASS  webhooks/webhooks.test.ts
  verifyTelegramWebhookSignature
    √ should fail closed when no secret token is configured
    √ should reject a request with no secret-token header
    √ should reject a mismatched secret token
    √ should accept a matching secret token
    √ should accept the header when provided as an array
  message handler
    √ rejects with 401 and persists nothing when no secret is configured
    √ rejects a forged secret token with 401 and persists nothing
    √ accepts a correctly signed update and persists the message
```

**Working proof:** the CI run for this branch executes that suite on your infrastructure rather than my machine — repo CI: https://github.com/corsairdev/corsair/actions

## Additional Notes

- Scope is `packages/telegram/**` only (R1).
- **Behaviour-changing for live traffic**, and worth a release note: any deployment currently running Telegram webhooks *without* a `secret_token` will start getting 401s until one is set via `setWebhook`. That is the intent, but unlike the other fixes in this family it will bite working (if insecure) installs.
- Out of scope but noted while in the file: the check is `providedToken === secretToken`, a plain comparison of a shared secret rather than a timing-safe one — same observation as on #609 for GitLab. Happy to send that separately.
- This was the last instance of the fail-open family I could find; #626/#627 cover a different root cause (empty `value[]`) in the Graph plugins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Telegram webhook requests are now rejected when no webhook secret is configured.
  * Invalid, missing, or unauthorized signature headers now consistently return an unauthorized response.
  * Unauthorized webhook requests no longer create or persist messages.
  * Valid requests continue to be accepted, including supported string and array header formats.

* **Tests**
  * Added comprehensive coverage for webhook authentication and message handling scenarios.

### DIFF
--- a/packages/telegram/webhooks/webhooks.test.ts
+++ b/packages/telegram/webhooks/webhooks.test.ts
@@ -1,0 +1,140 @@
+import type { WebhookRequest } from 'corsair/core';
+import type { TelegramContext } from '../index';
+import { message } from './message';
+import { verifyTelegramWebhookSignature } from './types';
+
+const SECRET = 'telegram-secret-token';
+
+function makeRequest(
+	payload: unknown,
+	headers: Record<string, string | string[]> = {},
+): WebhookRequest<unknown> {
+	return {
+		payload,
+		rawBody: JSON.stringify(payload),
+		headers,
+	} as unknown as WebhookRequest<unknown>;
+}
+
+function makeCtx(key: string): TelegramContext {
+	return {
+		key,
+		db: {
+			messages: {
+				upsertByEntityId: jest.fn().mockResolvedValue({ id: 'entity-1' }),
+			},
+		},
+	} as unknown as TelegramContext;
+}
+
+const messageUpdate = {
+	update_id: 1,
+	message: {
+		message_id: 42,
+		date: 1747872000,
+		chat: { id: 1234, type: 'private' },
+		from: { id: 9999, is_bot: false, first_name: 'Mallory' },
+		text: 'a forged message',
+	},
+};
+
+describe('verifyTelegramWebhookSignature', () => {
+	it('should fail closed when no secret token is configured', () => {
+		// The regression: an unset secret returned { valid: true }, so any caller
+		// who knew the webhook URL was treated as Telegram.
+		const result = verifyTelegramWebhookSignature(
+			makeRequest(messageUpdate),
+			'',
+		);
+		expect(result).toEqual({ valid: false, error: 'Missing webhook secret' });
+	});
+
+	it('should reject a request with no secret-token header', () => {
+		const result = verifyTelegramWebhookSignature(
+			makeRequest(messageUpdate),
+			SECRET,
+		);
+		expect(result).toEqual({
+			valid: false,
+			error: 'Missing x-telegram-bot-api-secret-token header',
+		});
+	});
+
+	it('should reject a mismatched secret token', () => {
+		const result = verifyTelegramWebhookSignature(
+			makeRequest(messageUpdate, {
+				'x-telegram-bot-api-secret-token': 'wrong',
+			}),
+			SECRET,
+		);
+		expect(result).toEqual({ valid: false, error: 'Invalid secret token' });
+	});
+
+	it('should accept a matching secret token', () => {
+		const result = verifyTelegramWebhookSignature(
+			makeRequest(messageUpdate, {
+				'x-telegram-bot-api-secret-token': SECRET,
+			}),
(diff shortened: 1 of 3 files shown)


packages/telegram/webhooks/webhooks.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+function makeCtx(key: string): TelegramContext {
+	return {
+		key,
+		db: {
+			messages: {
+				upsertByEntityId: jest.fn().mockResolvedValue({ id: 'entity-1' }),
+			},
+		},
+	} as unknown as TelegramContext;
```
ambikeesshh, line 27: can we add `$getAccountId` to this mock? the valid-path test currently catches a `ctx.$getAccountId is not a function` error and logs a warning.

ambikeesshh (commented):
fix looks good. please address the inline note, then this should be good to go
thanks!

ambikeesshh:
@greptileai review

ambikeesshh (commented):
all findings are resolved
lgtm

devjain32 (approved):

PR #658, opened 2026-08-10
fix(telegram): webhook verification fails open when no secret token i…

### LINKED ISSUE #628: fix(telegram): webhook verification fails open when no secret token is configured, allowing forged messages to be persisted

CaesarCipher2006:
…s configured, allowing forged messages to be persisted

Fixes #628

## Checklist

- [ ] I have run `pnpm lint` and all checks pass
- [ ] I have run `pnpm typecheck` and there are no TypeScript errors
- [ ] I have run `pnpm build` and all packages build successfully
- [ ] I have run `pnpm test` and all tests pass
- [ ] I have added or updated tests where applicable
- [ ] I have added or updated necessary documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook requests are now rejected when no verification secret is configured.
  * Improved validation responses for missing secrets, missing headers, valid secrets, and invalid tokens.

* **Tests**
  * Added coverage for webhook signature verification scenarios and expected error messages.

### DIFF
--- a/packages/telegram/webhooks/types.test.ts
+++ b/packages/telegram/webhooks/types.test.ts
@@ -1,0 +1,58 @@
+import { verifyTelegramWebhookSignature } from './types';
+
+describe('verifyTelegramWebhookSignature', () => {
+    const request = {
+        headers: {
+            'x-telegram-bot-api-secret-token': 'test-secret',
+        },
+    };
+
+    it('rejects when webhook secret is missing', () => {
+        const result = verifyTelegramWebhookSignature(
+            request as any,
+            '',
+        );
+
+        expect(result).toEqual({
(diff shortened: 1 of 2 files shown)


Dhirenderchoudhary:
Thanks for Contribution @CaesarCipher2006  Closing as duplicate fixed by #629.

PR #731, opened 2026-08-13
feat(harvest): add Harvest integration (57 ops)

### LINKED ISSUE #730: Integration request: Harvest

Harvest

### API documentation link

https://help.getharvest.com/api-v2/

Harvest is a time-tracking and invoicing tool for teams and freelancers. I want
Corsair agents to be able to log billable hours, manage the projects and clients
those hours belong to, and drive the invoicing that follows from them.

Specifically I need the 57 operations listed for Harvest on corsair.dev/oss:

- Clients (5) and client contacts (4)
- Company settings (2)
- Projects (5) and tasks (5)
- Time entries (5)
- Users (5)
- Expenses and expense categories (3)
- Invoices, invoice messages, payments and item categories (14)
- Estimates, estimate messages and item categories (9)

I'm doing this so an agent can answer "how many billable hours went to this
client last month" and act on the answer - raise the invoice, record the
payment, close the invoice - without a human moving data between the API and a
spreadsheet.

_No response_

### Additional context

Auth: OAuth 2.0. Harvest sends the credential as `Authorization: Bearer <token>`
alongside a `Harvest-Account-Id` header, because one token can reach several
accounts, so the account id is a second credential rather than something the
token implies. A personal access token is byte-compatible with an OAuth access
token, which makes local development and integration testing straightforward.

Harvest also requires a `User-Agent` header on every request and answers 400
Bad Request without one.

Rate limits: 100 requests per 15 seconds for the general API. The Reports API
has a much stricter 100 per 15 minutes, but no reporting operation is in the
catalog's 57, so a single rate-limit configuration covers everything here.
Successful responses carry no rate-limit headers at all, so a client can only
react to the 429 and its `Retry-After` rather than pacing itself in advance.

Errors use ordinary HTTP status codes (403, 404, 422, 429, 500), so the standard
error-handler mapping applies without inspecting response bodies.

Every list endpoint shares one pagination envelope
({ <resource>: [...], page, per_page, total_pages, total_entries, next_page,
previous_page, links }), so all 20 list operations can share a single schema.

Docs: https://help.getharvest.com/api-v2/

Agam00:
## Description

Adds the Harvest integration: all 57 operations listed for `harvest` on
corsair.dev/oss, covering time tracking, project and client management, expenses,
invoicing and estimates.

Fixes #730

Harvest API v2 docs: https://help.getharvest.com/api-v2/

### Operations (57)

| Group | Ops | Coverage |
| --- | --- | --- |
| Clients | 5 | list, get, create, update, delete |
| Client contacts | 4 | list, create, update, delete |
| Company | 2 | get, update |
| Projects | 5 | list, get, create, update, delete |
| Tasks | 5 | list, get, create, update, delete |
| Time entries | 5 | list, get, create, update, delete |
| Users | 5 | list, get, create, update, delete |
| Expenses | 3 | create, update, list categories |
| Invoices | 14 | 5 invoice + 3 message + 3 payment + 3 item category |
| Estimates | 9 | 4 estimate + 3 message + 2 item category |
| **Total** | **57** | |

The catalog surface is deliberately asymmetric - there is no List Estimates, no
Get/List/Delete Expense, no Get Client Contact and no Roles resource - and this
PR matches it exactly rather than filling the gaps, to stay inside R1 scope.

0 triggers: Harvest API v2 has no webhook, callback or streaming mechanism, so
`HarvestWebhooks` is `Record<string, never>` and `pluginWebhookMatcher` returns
false.

### Auth

`authType: PickAuth<'oauth_2'>`, matching the catalog. The runtime supplies the
access token in `ctx.key` and the client sends
`Authorization: Bearer <token>`.

Harvest needs a second credential: `Harvest-Account-Id`. One token can reach
several accounts, so the account is not implied by the token. It is handled the
same way `packages/zendesk` handles `subdomain`:

```ts
const accountId =
  ctx.options.accountId ?? (await ctx.keys.get_account_id()) ?? discover();
```

Discovery is a last resort via `https://id.getharvest.com/api/v2/accounts`,
filtered on `product === 'harvest'` because Forecast accounts appear in the same
list and are rejected by the Harvest API. It only resolves when the token can
reach exactly one Harvest account; with several it raises rather than guessing.

`User-Agent` is set inside the client rather than left to callers, because
Harvest answers 400 Bad Request to any request without one.

### Rate limiting

100 requests per 15 seconds, retried through `RateLimitConfig` with
`Retry-After` honoured and exponential backoff. Successful responses carry no
rate-limit headers at all - no `X-RateLimit-Remaining` to pace against - so the
client is necessarily reactive. This is documented in `client.ts` so the absence
does not read as an oversight.

The Reports API has a far stricter budget (100 per 15 minutes), but no reporting
operation is in scope, so one configuration covers every request the plugin
makes.

### Shared pagination envelope

Every Harvest list endpoint returns the same wrapper with only the data key
changing, so it is declared once and reused by all 20 list operations:

```ts
const withPagination = <Shape extends z.ZodRawShape>(shape: Shape) =>
  z.object({ ...shape, ...PaginationFields }).loose();
```

### Persistence - 10 entities

`clients`, `contacts`, `projects`, `tasks`, `users`, `invoices`, `estimates`,
`expenseCategories`, `invoiceItemCategories`, `company`.

Split on whether the data is reference or transactional. Reference data is
mirrored; time entries, expenses, invoice and estimate messages and payments are
not, because they are appended continuously and are only meaningful against a
date range.

Company settings are a per-account singleton with no `id`, so that entity is
keyed on `full_domain`.

Records are validated against the entity schema before being written, so a row
in an unrecognised shape is skipped rather than stored as something the plugin
cannot read back.

### Schemas built from live responses

Every entity field was captured from a real Harvest response (2026-08-13) rather
than transcribed from the docs - 216 fields across 13 resources. Only the primary
key is required; everything else is nullable and optional, and every schema is
`.loose()`.

That is deliberate: Harvest omits or nulls most fields depending on plan,
permissions and enabled features, so a stricter schema would reject valid rows.
`schema.test.ts` asserts both directions - every captured key is declared, and a
record carrying only its key still parses.

### Privacy

Harvest data is unusually personal for a plugin - contact email addresses and
phone numbers, invoice notes, time-entry descriptions of client work. Only
explicitly named identifier fields reach `corsair_events`; the names of the
remaining supplied fields are recorded without their values. Four tests assert
the exact event payloads, including that an invoice message logs the recipient
count rather than the addresses.

Two operations send real email and are documented as such: `users.create` sends
an invitation, and `invoices.createMessage` / `estimates.createMessage` with
`event_type: 'send'` mails the client. Neither is exercised by the live tests.

## Checklist

image

### Verification

```
lint (biome)        clean
typecheck           clean
validate:plugins    SUCCESS
tests               102 passed, 10 skipped (112 total)
build (tsup)        success
```

The 10 skipped tests are `integration.test.ts`, which self-skips unless
`HARVEST_ACCESS_TOKEN` and `HARVEST_ACCOUNT_ID` are set. They were run against a
real Harvest account during development and all 10 passed - company settings,
clients, projects, tasks, users, time entries, expense categories, invoices,
invoice item categories and pagination all parsed against the declared schemas.
Every one of them is read-only.

`endpoints.test.ts` includes a coverage sweep asserting that the set of
operations it exercises is exactly the set the plugin registers, so an operation
cannot be added without a test.

### Scope

30 files: 28 under `packages/harvest/`, the 3-line registration in
`packages/corsair/core/constants.ts` (+3/-0), and `pnpm-lock.yaml`. Nothing else.

### Response shapes verified against docs rather than live data

Six operations - invoice messages (3), estimate messages (2 of 3) and invoice
payments (3, list/create/delete) - have their routing, request bodies and event
payloads tested, but their response schemas come from the Harvest documentation
rather than a captured response, because creating those records is the one thing
in this API that can send mail to a client.

They are written fully optional and `.loose()`, so the risk is a field missing
from a schema rather than a valid response being rejected. Happy to seed and
verify them on request.

### Suggestion for core (not changed here, R1)

`SENSITIVE_QUERY_PARAMS` in `packages/corsair/async-core/ApiError.ts` does not
include `apikey`. It does not affect this plugin - Harvest authenticates by
header, not query string - but it is worth noting from the Alpha Vantage work,
where a query-string key could reach an `ApiError`.

## Summary by CodeRabbit

* **New Features**
  * Added Harvest as a supported integration.
  * Added secure authentication, account discovery, rate-limit retries, and error handling.
  * Added support for clients, contacts, company settings, projects, tasks, time entries, users, expenses, invoices, and estimates.
  * Added local caching, sanitized activity logging, request validation, pagination, and account selection.
* **Tests**
  * Added comprehensive coverage for Harvest operations, schemas, error handling, caching, pagination, and optional live integration scenarios.


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
Checked against the Harvest docs and a live account. Catalog’s right.

One real bug: createPayment left send_thank_you unset, Harvest defaults it to true and can email the client. We send false unless asked.

Also filled in the official fields we were dropping (contact recipient status, saml_exempt, invoice retainer/payment_options, time-entry approval/external_reference). 500s retry except on creates.

Live CRUD passed. Good to merge.

devjain32 (approved):

Issue #746, opened 2026-08-13
a11y(studio): remaining form controls have no accessible name


yashksaini-coder:
## What's wrong

Follow-up to #717 / #740 / #742. Those labelled the Chat and Database controls one at a time. Sweeping the rest of `packages/studio/src/web`, **8 interactive form controls still have no accessible name**, plus 2 disabled ones.

The largest cluster has a single shared cause. `FieldLabel` in `components/OperationInputForm.tsx:82` renders

```tsx
<label className="block text-xs …">{name}…</label>
```

as a **sibling** of the control (`FieldRenderer`, line 307 vs 308-313) — no `htmlFor`, and it does not wrap the input. So it contributes nothing to the accessible name, and every generated operation-input field is announced unnamed. Fixing that one component covers 4 of the 8.

## Affected controls

Unnamed and focusable:

| File | Line | Control | Current fallback |
| --- | --- | --- | --- |
| `components/OperationInputForm.tsx` | 111 | `<select>` (enum field) | none |
| `components/OperationInputForm.tsx` | 126 | `<Input>` (string field) | none |
| `components/OperationInputForm.tsx` | 146 | `<Input>` (number field) | none |
| `components/OperationInputForm.tsx` | 430 | `<Textarea>` (JSON editor) | none |
| `pages/ScriptPage.tsx` | 81 | `<select>` (tenant picker) | none |
| `pages/ScriptPage.tsx` | 109 | `<Textarea>` (script editor) | none |
| `pages/PluginDetail.tsx` | 193 | `<Input>` (OAuth code) | placeholder only |
| `pages/PluginsPage.tsx` | 244 | `<Input>` (create tenant) | placeholder only |

Lower priority — `disabled`, so not focusable: `OperationInputForm.tsx:334` and `:342`.

The two placeholder-only cases are the same class as #717: placeholder is the last-resort fallback in the [AccName spec](https://w3c.github.io/accname/#step2E) and is applied inconsistently across screen readers.

Not affected (already correct, listing so the sweep isn't redone): `PluginDetail.tsx:289` is implicitly labelled by a wrapping `<label>`.

## Suggested change

Two independent pieces, happy to split them:

1. `OperationInputForm` — give `FieldLabel` an `htmlFor` and the control a matching generated `id` (React's `useId`), so the visible label does the naming properly. Covers lines 111/126/146 and the array/JSON editors.
2. The four one-off sites — add `aria-label`, matching the approach in #740 / #742.

[...]

PR #742, opened 2026-08-13
fix(studio): add aria labels for selects

yuvanvk:
## Description

Adds accessible names to the tenant `<select>` and sort-field `<select>` on the Database page.
- Tenant `<select>`: associated with the visible "Tenant" label via  `aria-label="Tenant"`
- Sort field `<select>`: `aria-label="Sort field"`

No visual or behavioral changes — labeling only. Search input labeling is a separate issue and is not addressed here.

Fixes #695

## Checklist
- [x] tenant select has an accessible name
- [x] sort select has an accessible name
- [x] no visual redesign required

## Screenshots / Demos (if applicable)

`137:   aria-label="Tenant"`
`162:   aria-label="Sort field"`

## Additional Notes
Verified with:
`rg 'aria-label="Tenant"|aria-label="Sort field"' packages/studio/src/web/pages/DatabasePage.tsx`

## Summary by CodeRabbit

* **Accessibility**
  * Added descriptive labels to the tenant and sort-field selectors on the Database page, improving support for screen readers and other assistive technologies.

PR #740, opened 2026-08-13
fix(studio): add accessible labels to database search and sort direction toggle

akshatmalik-bruh:
## Description

Adds `aria-label="Search records"` to the search input and `aria-label="Sort direction"` to the ASC/DESC sort direction toggle button in Studio (`packages/studio/src/web/pages/DatabasePage.tsx`). 

These interactive elements previously lacked accessible names for screen readers.
fixes #719 

## Checklist

- [x] I have run `pnpm lint` and all checks pass
- [x] I have run `pnpm typecheck` and there are no TypeScript errors
- [x] I have run `pnpm build` and all packages build successfully
- [x] I have run `pnpm test` and all tests pass
- [ ] I have added or updated tests where applicable
- [x] I have added or updated necessary documentation

## Screenshots / Demos (if applicable)

image
image

## Additional Notes

- Select controls were left unchanged.
- No visual CSS styling modifications were made.
- Verified via `rg 'aria-label="Search records"|aria-label="Sort direction' packages/studio/src/web/pages/DatabasePage.tsx`.

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Added accessible labels to the database record search field and sort-direction control, including the current sort direction for clearer screen reader support.

PR #747, opened 2026-08-13
fix(studio): label the ChatPage message textarea

### LINKED ISSUE #717: a11y(studio): label the ChatPage message textarea
## What's wrong

The conversation select and Send button on Chat already have names. The composer textarea in `packages/studio/src/web/pages/ChatPage.tsx` still only has a placeholder.

## Change

Add `aria-label="Chat message"` on the Textarea. Do not restyle it.

## Check

`rg 'aria-label="Chat message"' packages/studio/src/web/pages/ChatPage.tsx`

yashksaini-coder:
## Description

Fixes #717.

The Chat composer relied on its `placeholder` for an accessible name. Placeholder text is only the last-resort fallback in the [AccName spec](https://w3c.github.io/accname/#step2E) and is applied inconsistently across screen readers, so the field had no stable name.

Adds `aria-label="Chat message"` to the `Textarea` in `packages/studio/src/web/pages/ChatPage.tsx`, matching the existing labels on the conversation `select` ("Active chat") and the Send button — same approach as #740 and #742. One line, no styling changes.

`Textarea` (`components/Primitives.tsx:44`) spreads `...rest` onto the native `<textarea>` and is typed as `React.TextareaHTMLAttributes<HTMLTextAreaElement>`, so the attribute reaches the DOM with no change to the primitive.

## Verification

Checked against the live studio UI in Chromium via the browser accessibility tree:

| | Accessible name of the composer |
| --- | --- |
| before | `textbox "Ask about your integrations… (Ctrl+Enter to send)"` |
| after | `textbox "Chat message"` |

No page errors, and no other control on the page changed.

## Checklist

- [x] I have run `pnpm lint` and all checks pass — exit 0 (6 pre-existing warnings elsewhere, none in this file)
- [x] I have run `pnpm typecheck` and there are no TypeScript errors — exit 0
- [x] I have run `pnpm build` and all packages build successfully — exit 0, 128/128 tasks
- [ ] I have run `pnpm test` and all tests pass — **see note below**
- [ ] I have added or updated tests where applicable — n/a, see note
- [ ] I have added or updated necessary documentation — n/a, no user-facing behaviour change

**On `pnpm test`:** it does not pass in my local environment, and does not pass on a clean `main` either — identical results both ways (17 suites / 103 tests failing). Every failure is `Could not locate the bindings file … better_sqlite3.node` in `packages/corsair`: `better-sqlite3` cannot compile against my local Node 26 because its C++ uses V8 APIs that were removed (`Object::GetPrototype`, `Context::GetIsolate`). This is unrelated to the change here, which touches only `packages/studio` web UI. Happy to re-run on Node 22 if you'd like that confirmed.

**On tests for this change:** `packages/studio` currently has no test setup (no test files, no vitest/testing-library/jsdom), so there is nothing to extend. Adding a test framework felt well outside the scope of a one-line a11y fix, and matches how #740 and #742 shipped. I verified via the browser accessibility tree instead, as above.

## Additional Notes

No breaking changes, no new dependencies.

While checking for sibling instances of this bug I found that 8 further form controls in `packages/studio/src/web` still have no accessible name — most of them sharing one root cause in `OperationInputForm`, where `FieldLabel` renders a `<label>` as a sibling of the input with no `htmlFor`. I've written that up separately as #746 rather than widening this PR, and I'd be glad to take it on.

## Summary by CodeRabbit

* **Accessibility**
  * Added an accessible label to the chat message input for improved screen reader support.

### DIFF
--- a/packages/studio/src/web/pages/ChatPage.tsx
+++ b/packages/studio/src/web/pages/ChatPage.tsx
@@ -559,6 +559,7 @@ export function ChatPage({ tenant }: { tenant: string }) {
 							value={input}
 							onChange={(e) => setInput(e.target.value)}
 							onKeyDown={onKeyDown}
+							aria-label="Chat message"
 							placeholder="Ask about your integrations… (Ctrl+Enter to send)"
 							disabled={streaming || chatLoading}
 						/>


yashksaini-coder:
@ambikeesshh can you review this please

ambikeesshh (commented):
thanks @yashksaini-coder

devjain32 (approved):

PR #751, opened 2026-08-13
 feat(loyverse): add Loyverse integration (59 ops)

### LINKED ISSUE #750: Integration request: Loyverse

Loyverse

### API documentation link

https://developer.loyverse.com/docs/

Loyverse is a free point-of-sale platform used by small retail and hospitality
businesses. Everything about the shop lives behind the till: what is on sale, what
it costs, what sold today, who bought it and what is left in stock. Getting any of
it out today means a human opening the back office and exporting a spreadsheet.

I would like an agent to be able to read and act on that directly, covering the 59
operations the catalog lists:

- Catalogue: items (4), item images (2), variants (4), categories (4),
  modifiers (4), discounts (4), taxes (4)
- People and partners: customers (4), suppliers (4)
- Account configuration: stores (2), employees (2), payment types (2),
  POS devices (4)
- Trading: receipts and refunds (4), shifts (2), inventory levels (2)
- Administration: webhook subscriptions (4), merchant information (1),
  OpenID Connect metadata (2)

The concrete case that motivated it: an agent can answer "which items sold below
their cost last week, and at which store" by reading receipts against the item and
variant catalogue, and then act on the answer by repricing the variant or adjusting
its per-store price - without anybody exporting anything.

_No response_

### Additional context

Base URL: https://api.loyverse.com/v1.0 - the version is in the path.

Auth: Authorization: Bearer <token>. Loyverse offers both OAuth 2.0 and personal
access tokens and presents them identically, so a personal access token is
byte-compatible with an OAuth access token. There is no second credential - no
account id, no subdomain - and no User-Agent requirement. Tokens are self-serve
from the back office, up to 20 per account.

Errors: real HTTP status codes with a consistent body,
{"errors":[{"code":"...","details":"...","field":"..."}]}. Observed live: 400, 401,
402, 404. Also documented: 403, 415, 429, 500. Authentication is checked before
routing, so an unknown path with no token returns 401 rather than 404.

402 PAYMENT_REQUIRED has two meanings and is worth knowing about: a lapsed
subscription, or a request that reached past a plan limit. The second is the one
that actually occurs on a free account - filtering receipts by a date more than 31
days back returns "Unable to retrieve receipts created earlier than 31 days ago.
Please subscribe to Unlimited sales history", while an unfiltered receipts list
still succeeds. Either way no retry helps.

Rate limits: documented as 300 requests per 300 seconds per account. Two things
worth flagging. Successful responses carry no rate-limit headers at all, so a
client can only react to a 429 rather than pace itself. And a 429 could not be
induced - 400 requests inside about 60 seconds were all answered normally - so the
documented limit does not appear to be enforced as written.

Pagination: one shared envelope across all 16 list endpoints,
{"<resource>": [...], "cursor": "<opaque>"}. The cursor key is omitted entirely on
the final page rather than set to null. limit defaults to 50 and is capped at 250,
enforced with a 400 rather than clamped. One asymmetry: GET /inventory returns its
array under inventory_levels, the only key that does not match its path segment.

Soft deletion: most resources soft-delete, and every deleted record drops out of
the default list while remaining retrievable with show_deleted=true and carrying
deleted_at. What a direct read returns afterwards is not uniform - 404 for items,
modifiers, taxes and POS devices, but 200 with deleted_at set for categories and
suppliers - so a 404 is not proof a record never existed. Customers are documented
as hard-deleted for personal-data reasons and carry permanent_deletion_at.

Query parameters are ignored rather than rejected when unrecognised, so a
misspelled filter returns 200 with the whole collection. The same applies to body
fields. Both are easy to mistake for a working call. Filter names are also
inconsistently pluralised: items_ids and variants_ids, but modifier_ids,
discount_ids and tax_ids.

Writes are upserts: a collection POST carrying an id updates that record, and
without one creates a new one, so there is one endpoint per resource rather than a
create/update pair.

The published OpenAPI document (loaded by the docs page as
API-Reference__v1.0.yaml) has several errors I found while building against it, all
verified live:

- JWKS is documented at /oidc/jwks, which 404s. The working location is the
  /.well-known/jwks.json that the discovery document advertises.
- The item request field is documented as modifiers_ids, but the API reads
  modifier_ids and silently ignores the documented name, so modifiers never attach.
- The webhook response returns a deleted_at that the spec does not declare.
- Creating a webhook requires status, though only url is listed as required.

Separately, the catalog's own webhook description names inventory.update and
receipts.create as event types; the API rejects both with 400. The spec's
inventory_levels.update and receipts.update are the accepted names.

Dates are UTC ISO 8601 with milliseconds. Ids are UUID strings everywhere except
receipts, which use a sequential string such as "0001".

Agam00:
## Description

Adds a Loyverse integration covering all 59 operations listed in the OSS
catalog: the full product catalogue (items, variants, categories, modifiers,
discounts, taxes), customers and suppliers, account configuration (stores,
employees, payment types, POS devices), inventory levels, sales and refunds,
shifts, webhook administration, merchant information, and the two OpenID Connect
metadata documents.

Loyverse is a point-of-sale platform, so the integration lets an agent answer
questions like "which items sold below cost last week" and act on the answer by
repricing the variant, without a human exporting a spreadsheet first.

Fixes #750 

Docs: https://developer.loyverse.com/docs/
Catalog: https://corsair.dev/oss/loyverse

## Operations

59 operations, matching the catalog exactly. 34 paths across 18 groups.

| Group | Ops | Endpoints |
| --- | --- | --- |
| Items | 4 | GET/POST `/items`, GET/DELETE `/items/{item_id}` |
| Item images | 2 | POST/DELETE `/items/{item_id}/image` |
| Variants | 4 | GET/POST `/variants`, GET/DELETE `/variants/{variant_id}` |
| Categories | 4 | GET/POST `/categories`, GET/DELETE `/categories/{category_id}` |
| Modifiers | 4 | GET/POST `/modifiers`, GET/DELETE `/modifiers/{modifier_id}` |
| Discounts | 5 | GET/POST `/discounts` (two list operations), GET/DELETE `/discounts/{discount_id}` |
| Taxes | 4 | GET/POST `/taxes`, GET/DELETE `/taxes/{tax_id}` |
| Customers | 4 | GET/POST `/customers`, GET/DELETE `/customers/{customer_id}` |
| Suppliers | 4 | GET/POST `/suppliers/`, GET/DELETE `/suppliers/{supplier_id}` |
| POS devices | 4 | GET/POST `/pos_devices`, GET/DELETE `/pos_devices/{pos_device_id}` |
| Webhooks | 4 | GET/POST `/webhooks/`, GET/DELETE `/webhooks/{webhook_id}` |
| Inventory | 2 | GET/POST `/inventory` |
| Employees | 2 | GET `/employees`, GET `/employees/{employee_id}` |
| Payment types | 2 | GET `/payment_types`, GET `/payment_types/{payment_type_id}` |
| Stores | 2 | GET `/stores`, GET `/stores/{store_id}` |
| Shifts | 1 | GET `/shifts` (list only - the catalog lists no Get Shift) |
| Receipts | 4 | GET/POST `/receipts`, GET `/receipts/{receipt_number}`, POST `/receipts/{receipt_number}/refund` |
| Merchant | 1 | GET `/merchant/` |
| OIDC | 2 | discovery document and JWKS |

Operations are named `upsert` rather than paired create and update, because
Loyverse exposes one endpoint for both: a collection POST carrying an `id`
updates that record, and without one creates a new record. This was verified
against the live API rather than inferred.

The surface was checked operation by operation against the catalog's operation
ids rather than by counting. Two consequences worth calling out, because the
totals match either way and a count alone would hide both:

- **The catalog lists two discount list operations.** Loyverse exposes only one
  `GET /discounts`, so `discounts.list` is the plain read and
  `discounts.listFiltered` is the same endpoint with the id filter, the four
  timestamp bounds and `show_deleted`. Both are registered so the surface matches.
- **There is no Get Shift.** The API does serve `GET /shifts/{shift_id}`, but the
  catalog lists only List Shifts, so only the list is implemented. The surface
  follows the catalog rather than the API.

Two other catalog labels are auto-generated names for ordinary endpoints and are
named plainly here: "List Employees (Advanced)" is `GET /employees` and "List
Modifiers (Enhanced)" is `GET /modifiers`.

Read-only by design, matching the API: employees, payment types, stores and
shifts have no write operations because Loyverse manages them in the back office
or on the till. That asymmetry is the API's, not an omission here.

### Filters

Every collection accepts its own id filter, and several accept more - `sku` on
variants, an exact `email` on customers, a store filter on POS devices, inventory
and shifts, and on receipts a range over the receipt number itself plus `order`
and `source`. All of them are exposed.

These were verified **by effect rather than by status code**, because Loyverse
ignores an unrecognised query parameter and answers 200 with the whole collection.
A misspelled filter name is therefore indistinguishable from a working call unless
you check that the result actually narrowed - `integration.test.ts` asks for one
known id per resource and fails if more than that row comes back. The API is also
inconsistent about pluralisation (`items_ids` and `variants_ids`, but
`modifier_ids`, `discount_ids` and `tax_ids`), so no name is derived from the
resource.

Only 10 of the 16 collections actually paginate. `/discounts`, `/taxes`,
`/payment_types`, `/stores`, `/pos_devices` and `/webhooks/` ignore a cursor and
return everything; the paginating ones reject an invalid cursor with 400. The
webhook collection takes no query parameters at all, so none are sent.

## Triggers

Zero, matching the catalog.

Loyverse does publish five webhook events, and the four `webhooks.*` operations
administer those subscriptions. The event names were taken from neither document
on trust: the catalog's prose names `inventory.update` and `receipts.create`, both
of which the API rejects with 400, while the spec's `inventory_levels.update` and
`receipts.update` are accepted. The enum ships the accepted five, and
`integration.test.ts` asserts the two rejected names still fail.

No Corsair trigger is registered, for a concrete
reason: a subscription created with a personal access token belongs to the
merchant and its notifications are **unsigned** - no `X-Loyverse-Signature`
header. Only subscriptions created through an OAuth application are signed, using
that application's client secret. Without a signature there is nothing to verify
a delivery against, so registering a trigger would mean accepting unauthenticated
POSTs.

## Authentication

`Authorization: Bearer <token>`.

Loyverse offers OAuth 2.0 and personal access tokens, and both are presented
identically, so a personal access token is byte-compatible with an OAuth access
token and the runtime-supplied key works for either. Development and the live
tests used a personal access token.

There is no second credential. Unlike Harvest's account id or Zendesk's
subdomain, the token alone identifies the account, so no `account` keys are
declared and there is no resolution chain or discovery fallback. No `User-Agent`
is required either.

## Rate limiting

Documented as 300 requests per 300 seconds per account, answered with 429
`RATE_LIMITED`.

Two limits of the implementation are worth stating rather than implying they were
measured:

- Successful responses carry **no** rate-limit headers of any kind - no
  remaining-quota header to pace against - so the client can only react to a 429
  rather than throttle proactively. Confirmed by inspecting the response headers
  of a successful call.
- **A 429 could not be induced.** 400 requests inside about 60 seconds were all
  answered normally, so the documented limit is not enforced as written. The
  `Retry-After` header name is therefore configured from the documented contract
  and has **not** been observed on a real throttle. The exponential backoff
  applies regardless of whether the header is present.

A rejected request was never applied, so the 429 handler retries regardless of
operation. That is the only handler that does.

## Pagination

One shared cursor envelope covers all 16 list operations, declared once:

```json
{ "<resource>": [ ... ], "cursor": "<opaque>" }
```

`cursor` is optional rather than nullable because Loyverse **omits the key
entirely** on the final page rather than sending `null` - verified by paging
through items with `limit=1`. `limit` defaults to 50 and is capped at 250, which
the API enforces with a 400 rather than clamping, so the bound is checked before
the request is made.

One asymmetry: `GET /inventory` returns its array under `inventory_levels`, the
only collection whose key does not match its path segment.

## Persistence

13 entities are mirrored: items, variants, categories, modifiers, discounts,
taxes, customers, suppliers, stores, employees, payment types, POS devices and
merchant.

These are reference data - the catalogue, who the customers and suppliers are,
and how the account is configured. They change rarely and are the lookup every
other operation needs. Reading an item also mirrors the variants nested inside
it, so a variant does not have to wait for a separate call.

**Not mirrored, deliberately:**

- **Receipts and refunds.** Transactional, appended on every sale, and only
  meaningful against a date range. The reference data a receipt points at is
  mirrored instead, which is what makes a receipt readable after the fact.
- **Shifts.** Transactional in the same way, appended as tills close.
- **Inventory levels.** These have no id at all - the natural key is
  `variant_id` plus `store_id` - and the figure changes on every sale, so a
  mirrored stock level would be stale by design and actively misleading.
- **Webhook subscriptions.** Account configuration, not business reference data.

**Eviction on delete.** All 9 deletes that have a mirrored entity call
`deleteByEntityId`; reads never evict. For customers this is a correctness
requirement rather than tidiness: Loyverse does not soft-delete customers - the
API documents this as a personal-data restriction and the record carries
`permanent_deletion_at` - so a row left behind would outlive a record that no
longer exists and would keep answering reads with personal data the account has
deleted. Deleting an item also evicts the variant ids the API reports alongside
it.

Everything else soft-deletes, though **not uniformly**, which is worth knowing
before treating a 404 as proof a record never existed. After a delete, a direct
read returns 404 for items, modifiers, taxes, customers and POS devices, but 200
with `deleted_at` set for categories and suppliers. What is consistent is that
every deleted record drops out of the default list and reappears under
`show_deleted=true` - that list is the view the mirror reflects, and it is why an
explicit delete evicts while a read does not. A stale row otherwise keeps its value
for resolving what a year-old receipt refers to. Measured per resource rather than
assumed, and pinned by a live test.

## Schemas

Built from live responses captured on 2026-08-13, not transcribed from the
documentation. `schema.test.ts` lists the field names observed per entity and
asserts each one is declared, so an entity cannot silently lose a field the API
returns.

Only the primary key is required on every entity; everything else is nullable and
optional, and every object is `.loose()`. Loyverse nulls or omits most fields
depending on which features an account has enabled, so a stricter schema would
reject valid rows, and a rejected row is a lost row. Note that variants are keyed
on `variant_id`, the one entity whose primary key is not `id`.

**One schema of 17 is not from a captured response.** `shifts` could not be
captured: a shift only exists once a cashier opens and closes one in the Loyverse
POS app, which needs a device login, and there is no endpoint to create one, so
the development account returned an empty collection. Those fields come from the
published spec. Two mitigations apply - every field is nullable and optional, as
everywhere else, and the object is `.loose()`, so a field that differs in name or
presence cannot reject a row.

### Corrections to the published specification

The plugin diverges from `API-Reference__v1.0.yaml` in three places, each verified
against the live API:

1. **JWKS path.** The spec documents `/oidc/jwks`, which returns 404. The working
   location is the `jwks_uri` the discovery document advertises,
   `/.well-known/jwks.json`. Implementing the spec verbatim would have shipped an
   operation that never worked.
2. **Item modifier field.** The spec documents the request field as
   `modifiers_ids`. The API accepts that name without complaint and then ignores
   it - the modifiers never attach. Both the request and the response use
   `modifier_ids`.
3. **Undeclared response field.** The live webhook response returns `deleted_at`,
   which the spec's `Webhook` schema does not declare.

A fourth correction is to the catalog rather than the spec: its webhook
description names two event types the API rejects, as described under Triggers.

Further constraints are absent from the documentation entirely and were found
live, so they are encoded in the input schemas with the reason:

- Updating an item without a `variants` array is rejected with
  `INVALID_VALUE  Could not update variants to []` - Loyverse reads an absent
  array as an instruction to remove every variant, so an update must restate the
  variants it keeps.
- A refund line must carry the `id` of the line on the original receipt; a
  `variant_id` alone returns
  `MISSING_REQUIRED_PARAMETER  field: object.line_items[0].id`.
- Creating a webhook without `status` returns
  `MISSING_REQUIRED_PARAMETER  field: object.status`, though the spec lists only
  `url` as required.
- Creating a variant requires the parent item to already declare option names.
  Without them Loyverse answers `INVALID_VALUE  Could not add variant to item with
  id '...' because it has no options`, and with them but no matching
  `option1_value` it answers `INVALID_VALUE  You cannot add or delete options for
  an existing item with variants`. The schema cannot see the parent item, so it
  enforces what it can - a create needs an `item_id` - and documents the rest on
  the field.

## Retry safety

Corsair replays the whole endpoint call when a handler asks for a retry, and
Loyverse accepts no idempotency key, so a network failure raised after a write
committed would apply it twice. 13 of the 14 POST operations are therefore never
retried on a network error.

`inventory.update` is the exception and is deliberately retryable: its body
carries `stock_after`, an absolute level rather than a delta, so applying it twice
leaves the same figure. Verified live - two identical calls with `stock_after: 7`
both left the level at 7.

The `upsert` operations are the subtle case. They are idempotent when an `id` is
supplied and create a duplicate when it is not, and the predicate only sees the
operation name, so it reports every upsert as unsafe. That costs a retry which
would sometimes have been fine and never creates a duplicate record.

Deletes are retryable: repeating one is not a duplication risk, and the second
attempt reports 404, which the not-found handler already covers.

`endpoints.test.ts` asserts this set equals exactly the POST operations minus
`inventory.update`, so the predicate cannot drift from the endpoints it describes.

`402 PAYMENT_REQUIRED` is matched before the generic handlers and never retried.
It has two distinct meanings and the handler names both rather than guessing: a
lapsed subscription, or a request that reached past a plan limit. The second is the
one that actually occurs on a free account - filtering receipts by a date more than
31 days back returns `Unable to retrieve receipts created earlier than 31 days ago.
Please subscribe to Unlimited sales history`, while an unfiltered receipts list
still succeeds. An earlier draft reported it only as a lapsed subscription, which
would have sent an operator to the billing page over a query they could simply
narrow.

## Privacy

Audit payloads carry named identifiers and counts only.

`customers` is the sensitive entity: its 20 live fields include name, email,
phone number, postal address, a free-text note and lifetime spend. Employees
carry a name and email, suppliers carry contact details, and receipts carry a
free-text `note` and per-line `line_note` written at the till.

None of it reaches `corsair_events`:

- A customer write logs the customer id and whether it was a create. The generic
  audit helper is deliberately not used there, so not even the field *names*
  suggest the log holds contact details.
- A receipt logs its number, store, line count and payment count - never line
  contents or notes.
- An image upload logs a byte count, never the image.
- A modifier logs an option count, not the option names.
- A webhook logs type and status, not the URL, which can carry a token.

`endpoints.test.ts` asserts the exact event payload for each of these and
additionally asserts that no personal value or field name appears in the
serialised payload.

## Checklist

- [x] One plugin per PR
- [x] Tests with real assertions
- [x] Description complete
- [x] Linked issue
- [x] Screenshot of the test run
- [x] No secrets, no `eval`, no `new Function`
- [x] No `any` on exported surfaces
- [x] Registered in `packages/corsair/core/constants.ts`
- [x] `dist/` not committed

## Screenshots / Demos

image

## Verification

Run on Node 22.18.0. CI runs Node 24, so these are a proxy rather than proof.

| Check | Result |
| --- | --- |
| `pnpm lint` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm run validate:plugins` | SUCCESS |
| `pnpm run validate:docs` | SUCCESS |
| `pnpm build` (packages/loyverse) | exit 0, 47.87 KB ESM |
| `npx jest --passWithNoTests --ci --testPathIgnorePatterns="api\.test\.ts\|integration\.test\.ts"` | 3 suites, 197 tests passed |
| `integration.test.ts` against a live account | 30 tests passed |

Test files: 4. Assertions: 204.

The operation surface is additionally checked against the catalog's operation ids
one by one rather than by total, since both a missing and a surplus operation can
leave the count sitting at 59.

`integration.test.ts` is excluded from CI and self-skips without
`LOYVERSE_ACCESS_TOKEN`, so a checkout with no credentials still runs green. It
never modifies anything that was already on the account. One test does write - it
creates two records it owns, named as probes, and deletes them again, because the
non-uniform soft-delete behaviour it pins cannot be observed any other way. The two
receipt operations and the image upload are deliberately never run live: the first
two move money and cannot be withdrawn, and the third would alter a real product
listing. Its authenticated tests
parse real rows against every entity schema, prove each id filter genuinely
narrows its result, and confirm the two rejected webhook event names still fail;
three further tests run unauthenticated against the OIDC documents.

## Scope

38 files:

- 36 new files under `packages/loyverse/`
- `packages/corsair/core/constants.ts`, exactly +3/-0
- `pnpm-lock.yaml`

## Known limitations

- The `shifts` response schema comes from the published spec rather than a
  captured response, for the reason given above.
- A 429 was never observed, so the `Retry-After` handling is unverified against a
  real throttle.
- Composite item `components` were never observed populated - the feature was not
  enabled on the development account - so that nested shape comes from the spec.
- The item image upload rejects a very small image with a 500 rather than a 4xx: a
  1x1 PNG fails where a 64x64 one succeeds. The input schema enforces a minimum
  payload length to catch an empty or truncated image early, but a small valid
  image can still produce an unhelpful server error.
- The upload is a **raw binary body**, not multipart - a `multipart/form-data`
  request is answered with 500 INTERNAL_ERROR. Both media types the catalog names
  are accepted, and the API does not appear to validate the declared type against
  the bytes: PNG bytes labelled `image/jpeg`, and a request with no content type at
  all, were both accepted with 201. The caller still declares one.
- `shifts` is list-only here because the catalog lists no Get Shift, even though
  `GET /shifts/{shift_id}` exists.

## Core suggestions, deliberately not implemented

Out of scope for a plugin PR under R1, so noted here instead:

`SENSITIVE_QUERY_PARAMS` in `packages/corsair/async-core/ApiError.ts:4` currently
holds `['api_key', 'key', 'token', 'appid']`. Loyverse carries its credential in
the `Authorization` header, so nothing leaks through a query string here and this
plugin is unaffected. It is worth noting only because `cursor` values are opaque
and appear in URLs, and an `ApiError` message includes the URL - harmless for
Loyverse, but a provider that put a session token in a query parameter under a
different name would leak it into an error message.

## Summary by CodeRabbit

* **New Features**
  * Added Loyverse as a supported provider.
  * Added OAuth-authenticated access to catalog, customers, inventory, sales, payments, stores, employees, webhooks, and account configuration.
  * Added item image upload and deletion.
  * Added local synchronization for supported Loyverse records.
  * Added OpenID Connect discovery and signing-key retrieval.
* **Bug Fixes**
  * Added rate-limit handling, retries, and operation-specific error handling.
* **Tests**
  * Added comprehensive mocked, schema, endpoint, and live integration coverage.


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
Labeled the persisted schema fields from the official Loyverse docs. Live keys match; tests pass.
LGTM

PR was good thanks @Agam00

devjain32 (approved):

PR #752, opened 2026-08-13
feat: add ActiveCampaign plugin with endpoints, schema, and tests

### LINKED ISSUE #749: integration request: ActiveCampaign

ActiveCampaign

### API documentation link

https://developers.activecampaign.com/reference/overview

I want an agent to run marketing and sales operations against an ActiveCampaign
account end to end, rather than only reading from it.

ActiveCampaign is unusual in that a single account holds four systems that are
normally separate products:

- the **contact database** (contacts, custom fields, lists, tags, segments)
- the **CRM** (deals, pipelines, stages, tasks, accounts)
- the **campaign and messaging layer** (campaigns, messages, templates, automations)
- an **e-commerce data store** (orders, products, customers, recurring payments)

Today an agent that needs to answer a question spanning two of them has to read
from one place and act in another. I want both halves in one plugin.

### Scope

298 operations across 30 resource groups:

| Domain | Ops |
| --- | --- |
| Contacts, custom fields, lists, tags, notes | 69 |
| Deals, pipelines, stages, deal tasks and fields | 46 |
| Campaigns, messaging, templates, automations | 40 |
| E-commerce (REST + GraphQL + browse sessions) | 34 |
| Accounts and account custom fields | 24 |
| Segments and audiences | 16 |
| Account configuration (addresses, calendars, branding) | 16 |
| Site and event tracking | 14 |
| Users, groups, permissions | 14 |
| Custom objects and schemas | 11 |
| SMS broadcasts and metrics | 8 |
| Webhook management | 6 |
| **Total** | **298** |

### Example workflow this enables

An order arrives through the e-commerce connection, and an agent:

1. looks up the contact by email
2. checks which lists and automations they are already enrolled in
3. adds a note to the associated deal
4. moves the deal to the next pipeline stage
5. tags the contact for follow-up

That is five operations across four ActiveCampaign subsystems, and today it is
four screens for a human.

No Corsair triggers are planned for this integration.

ActiveCampaign does offer webhooks, and this plugin exposes six operations for
managing webhook subscriptions via `/api/3/webhooks`:

- create a webhook
- retrieve a webhook
- update a webhook
- delete a webhook
- list webhooks
- list available webhook events

These are management operations against the ActiveCampaign API, not Corsair
trigger definitions. No inbound webhook receiver and no tenant-matching
infrastructure is part of this plugin.

### Authentication

API key sent in an `Api-Token` HTTP header. There is no OAuth flow to run.

A **second credential is required**, because the base URL is account-specific:

    https://<account>.api-us1.com/api/3/

The account subdomain cannot be derived from the API key, so it has to be
supplied alongside it.

| Credential | Where it goes | Derivable? |
| --- | --- | --- |
| API key | `Api-Token` request header | n/a |
| Account subdomain | Base URL host | No - must be supplied |

This matches the pattern the Zendesk plugin already uses in this repo, which
takes a subdomain the same way.

### Two APIs behind one credential

| Surface | Path | Covers |
| --- | --- | --- |
| v3 REST | `/api/3/...` | Contacts, deals, campaigns, lists, tags, users, segments, custom objects |
| eComm GraphQL | `/api/3/ecom/graphql` | Products, bulk order upsert (sync + async), recurring payments, browse sessions |

Both sit on the same base URL and use the same `Api-Token` header, and both
share one rate limit. The plugin will use a single transport that can issue
either, rather than two separate clients.

### Rate limits

**5 requests per second per account**, shared across REST and GraphQL.
Exceeding it returns `429`.

Unlike some APIs, ActiveCampaign sends rate-limit headers on _successful_
responses, not only on rejections:

| Header | Meaning |
| --- | --- |
| `Retry-After` | Seconds to pause after a 429 |
| `RateLimit-Limit` | Maximum requests allowed in the window |
| `RateLimit-Remaining` | Requests still available |

This means the client can throttle proactively rather than only reacting to a
429. The plugin's rate-limit config will read `Retry-After`.

### Pagination

The v3 REST list endpoints share one envelope, so it will be declared once and
reused across every list operation:

- `limit` and `offset` query parameters
- collection under a resource-named key
- count under `meta.total`
- default page size **20**, maximum **100**

One documented exception: for the **contacts** collection, ActiveCampaign
recommends `id_greater` with `orders[id]=ASC` instead of offset paging, because
offset degrades on large contact lists. The contact list operation will support
both.

### Error handling

Conventional HTTP status codes for error states, so the standard error-handler
mapping works without inspecting response bodies. Exact per-operation codes
will be confirmed against live responses during implementation rather than
transcribed from the docs.

### Scope note

At 298 operations this is a large surface - roughly five times a typical
plugin. If maintainers would prefer this split across multiple PRs by resource
group, I am happy to do that; say which grouping you want and I will follow it.

abhishek-2k23:
## Description

Adds an ActiveCampaign integration covering the full catalog surface: contacts,
custom fields, lists, tags, the CRM, campaigns and messaging, automations,
segments, e-commerce, custom objects, tracking, SMS and account administration.

ActiveCampaign is unusual in that one account holds four systems that are
normally separate products - the contact database, the CRM, the campaign layer,
and an e-commerce store. An agent answering a question that spans two of them
previously had to read from one place and act in another. This plugin puts both
halves behind one credential.

Fixes #749
API documentation: https://developers.activecampaign.com/reference/overview

### Coverage

**304 operations across 67 resource groups**, implementing all **298 rows** of
the OSS catalog. Six operations have no catalog row of their own and are
included because they fill real gaps in the REST surface: `contacts.get`,
`contactLists.list`, `contactTags.list`, `ecomOrders.find`, `ecomOrders.upsert`
and `ecomOrderProducts.listForOrder`.

Risk levels: **145 read, 113 write, 46 destructive**.

| Domain | Ops |
| --- | --- |
| Contacts, custom fields, lists, tags, notes | 71 |
| Deals, pipelines, stages, tasks, deal fields | 49 |
| Campaigns, messaging, templates, forms, variables | 40 |
| E-commerce, REST and GraphQL, browse sessions | 34 |
| Users, groups, addresses, calendars, branding | 29 |
| Accounts and account custom fields | 24 |
| Segments, legacy and V2 | 20 |
| Site and event tracking, webhooks | 17 |
| Custom objects and records | 12 |
| SMS broadcasts, lists, credits, metrics | 8 |
| **Total** | **304** |

### Authentication

API key in an `Api-Token` header. No OAuth flow.

A second credential is required, because the base URL is account-specific -
`https://<account>.api-us1.com/api/3` - and the account slug cannot be derived
from the key. It is declared as `api_key: { account: ['account'] }`, which
generates `ctx.keys.get_account()`. This mirrors how `packages/zendesk` takes a
subdomain.

Resolution happens in one shared helper, `endpoints/shared.ts#resolveAccount`,
which raises the core's `AuthMissingError` when the slug is absent rather than
returning an empty string - an empty slug would otherwise be interpolated into
the hostname and surface a configuration gap as a confusing transport failure
against `https://.api-us1.com`. The error handler routes `AuthMissingError` to
`CONFIGURATION_ERROR` so it is never retried.

The slug is validated against `^[a-zA-Z0-9-]+$` before interpolation, so a
value containing a slash or a dot cannot redirect a request to another host.

### Two API surfaces, one transport

The e-commerce catalog - products, bulk order upsert, recurring payments and
browse sessions - is GraphQL at `/ecom/graphql`, on the same host and behind
the same `Api-Token` header and rate limit as the v3 REST API. `client.ts`
exposes both through one config builder, so auth and throttling cannot drift.

The GraphQL surface answers **200 with an `errors` array** rather than an HTTP
error status, so a GraphQL-level failure would pass straight through the
status-based error handlers and return an empty result as if it had succeeded.
The transport raises on that explicitly.

### Rate limiting

The documentation states 5 requests per second per account. The account used
for development returned `RateLimit-Limit: 1000` on every response, so the
documented figure and the observed header disagree and the client hardcodes
neither: it retries against `Retry-After` with exponential backoff (5 retries,
1s initial, 2x multiplier).

ActiveCampaign sends `RateLimit-Limit` and `RateLimit-Remaining` on successful
responses as well as rejections. Proactive throttling from those headers is not
implemented; the client reacts to 429.

### Pagination

One envelope for every REST collection - `limit` and `offset`, rows under a
resource-named key, count under `meta.total` - declared once in
`endpoints/shared.ts` and reused everywhere. `limit` is clamped to the
documented maximum of 100 rather than letting the API silently cap it.

The contacts collection additionally supports `id_greater` with
`orders[id]=ASC`, which ActiveCampaign documents as the performant path on
large contact lists.

### Schemas

Built from responses captured against a live account on 2026-08-13, not
transcribed from the documentation. `schema.test.ts` asserts every captured key
is declared, for the 13 entities whose shapes could be captured.

Almost every scalar ActiveCampaign returns is a JSON string - ids (`"1"`),
counts (`"0"`), booleans (`"0"`/`"1"`). There is a pattern to the exceptions,
found by capture rather than by reading: the newer **camelCase-keyed**
resources return real JSON types. `groupMembers`, `dealCustomFieldMeta` and
`accountCustomFieldMeta` all return genuine numbers, so their ids are coerced
and the ambiguous fields are unions. A string-only schema would have rejected
every row those endpoints send.

Only the primary key is required. Every other field is `.nullable().optional()`
and every object is `.loose()`, because ActiveCampaign omits or nulls fields
depending on plan, permissions and enabled features - and a rejected row is a
lost row.

### Persistence

**43 entities mirrored**, all reference data. Deliberately not mirrored, because
they are transactional rather than reference: deal activities, email
activities, contact automations, e-commerce orders, order products and order
activities, campaign links, and the account config store. `schema.test.ts`
asserts none of those is registered.

Cache writes validate against the entity schema before writing, so an
unrecognised row is skipped rather than stored unreadable - and a skip warns,
because silence would turn a schema gap into a row that never appears. Writes
are best-effort and bounded to 16 concurrent.

Deletes evict; reads deliberately do not, because ActiveCampaign archives far
more often than it deletes. Two deletions cascade upstream and so cascade in
the mirror: deleting a custom field destroys every value stored against it, and
deleting a tag removes every contact-tag association. `evictChildren` finds
those by foreign key and evicts them, so the mirror cannot outlive the records
it describes.

### Privacy

Audit payloads allow-list identifiers and pagination. Every other supplied key
is recorded by *name* only, never by value, so contact emails, names, phone
numbers, note bodies, message content and custom field values do not reach
`corsair_events`. `contacts.find` logs only the match count because its input
is an email address; `fieldValues.setForContact` logs the contact and field ids
but never the value; bulk import logs a contact count. `endpoints.test.ts`
asserts the exact payload and that the serialised form contains none of the
personal values.

### Retry safety

Corsair replays the entire endpoint call when a handler requests a retry, and
ActiveCampaign offers no idempotency key, so a network error raised after a
committed write would duplicate it. `NON_IDEMPOTENT_OPERATIONS` lists the 159
state-changing operations explicitly - not by name pattern, so a newly added
operation cannot silently opt into retries - and `endpoints.test.ts` asserts
that set equals the non-read operations in the registry exactly, in both
directions. 429s are still retried for writes, because the request was rejected
rather than applied.

The registry is structured so `<group>.<leaf>` camel-cased is exactly the
operation key (`fieldValues.setForContact` -> `fieldValuesSetForContact`), and
a test asserts that holds for all 304 paths - the retry-safety check depends on
translating one into the other, and a mismatch would make it skip operations
silently rather than fail.

### Fail-safe defaults

Three booleans default to the provider's value when omitted, and the provider's
value sends mail to real people. Each is sent explicitly instead:

- `send_last_broadcast` on list creation defaults to `true` upstream, which
  mails the account's most recent broadcast to every new subscriber. Sent as
  `false`.
- `exclude_automations` on bulk import defaults to running every automation a
  list subscription triggers. Sent as `true`.
- `useDefaults` on field values is sent explicitly so the default-filling
  behaviour is the caller's decision.

### Tests

**1192 tests across 7 suites**, 158 assertions.

| Suite | Covers |
| --- | --- |
| `routing.test.ts` | every one of the 304 operations against a mocked transport: base URL, `Api-Token` header, token never in the query string, method consistent with risk level, no `undefined` interpolated into a path |
| `behaviour.test.ts` | write-body envelopes, `undefined` omitted rather than serialised, the fail-safe defaults, mirroring into the correct store, reads never evicting, cascade eviction, credential resolution |
| `endpoints.test.ts` | registry invariants, risk levels, the non-idempotent set, error-handler ordering, audit payload redaction |
| `schema.test.ts` | every captured key declared, key-only rows parse, no transactional entity mirrored, the numeric-resource exceptions |
| `persist.test.ts` | validate-before-write, warn on skip, best-effort writes, bounded concurrency, eviction |
| `client.test.ts` | header shape, account-slug validation, REST and GraphQL sharing auth |
| `segments-v2.test.ts` | keeps the unverified-route declaration in step with the registry |

Inputs for the routing suite are generated by walking each operation's own zod
schema rather than hand-written 304 times, so a schema change cannot leave a
stale fixture behind.

## Checklist

## Screenshots / Demos
image

### Verification

| Command | Result |
| --- | --- |
| `biome check` | clean, 32 files |
| `pnpm typecheck` | 0 errors |
| `pnpm run validate:plugins` | `[SUCCESS] All plugins passed structural validation!` |
| `pnpm run validate:docs` | `[SUCCESS] Docs validation passed!` |
| `tsc --build` + `tsup` | build success, `dist/index.js` 176.86 KB |
| `turbo test -- --ci --testPathIgnorePatterns="api\.test\.ts\|integration\.test\.ts"` | 1192 passed, 7 suites |

### Scope

32 files under `packages/activecampaign/`, plus exactly +3/-0 in
alphabetically between `abstract` and `activetrail`. `pnpm-lock.yaml` also
changes, because the workspace gained a package. `dist/` is gitignored and not
tracked.

### Known limitations

- **16 operations carry routes that could not be confirmed against a live
  account**: the 14 V2 segment operations, `imports.listAggregate` and
  `browseSessions.testEvent`. The development account answers 404 to
  `/v2/segments`, `/segments/v2`, `/api/v2/segments`, `/segments/{id}/counts`
  and `/segments/{id}/match`, while the legacy `/segments` collection answers
  200 on the same account - so the V2 surface appears to be plan-gated rather
  than misnamed, and the documentation pages were not reachable either. They
  are declared in `UNVERIFIED_ROUTES` in `endpoints/segments-v2.ts`, and
  `segments-v2.test.ts` keeps that list in step with the registry. Everything
  else was confirmed against live responses before being written.
- Row shapes for deals, pipelines, campaigns, automations and several other
  resources are declared from the documentation and marked as uncaptured in
  `schema/database.ts`, because a trial account holds no rows for them. The 13
  entities whose shapes were captured are asserted key-by-key.
- Five contact sub-resources and several sideloaded collections are typed
  `z.unknown()` where the envelope key is confirmed but the row shape was never
  observed. Typing them from a guess would be worse than declaring them
  unmodelled.
- Documented and observed rate limits disagree (5/sec documented,
  `RateLimit-Limit: 1000` observed). The client reacts to `Retry-After` rather
  than assuming either.
- No `integration.test.ts`. Live coverage is the recon capture the schemas were
  built from, plus the test run above.

`SENSITIVE_QUERY_PARAMS` in `packages/corsair/async-core/ApiError.ts` lists
`['api_key', 'key', 'token', 'appid']`. ActiveCampaign sends its credential in
a header, so nothing here leaks - but any provider whose key travels as a
differently named query parameter would have it appear in an `ApiError`
message. Flagging rather than fixing, since R1 confines this PR to the plugin.

## Summary by CodeRabbit

* **New Features**
  * Added ActiveCampaign integration with API-key authentication and account support.
  * Added broad coverage for contacts, accounts, deals, lists, tags, fields, content, imports, segments, e-commerce, tracking, SMS, webhooks, and custom objects.
  * Added REST and GraphQL operations with pagination, caching, audit logging, and schema validation.
  * Added retry handling for rate limits and transient read failures.

* **Bug Fixes**
  * Improved credential validation, error reporting, sensitive-data protection, and cache consistency.


Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
Checked Locally  Hit the live API, fixed schemas that didn’t match real responses, then cleaned up review nits: pagination, ID encoding, audit counts, and the bulk-import limit docs. Tests pass.

Dhirenderchoudhary:
This was a huge PR thanks for Contributing @abhishek-2k23 found few bugs only else PR was good

devjain32 (approved):

Issue #712, opened 2026-08-13
fix(vercel): fail closed when webhook secret is empty


ambikeesshh:
## What's wrong

`verifyVercelWebhookSignature` in `packages/vercel/webhooks/types.ts` will `createHmac('sha1', secret)` even if `secret` is `''`.

## Change

Add `if (!secret) return false;` before the HMAC. This helper returns a boolean, so no error string.

Leave the existing signature-header and `timingSafeEqual` logic alone.

## Check

Empty secret returns `false`. A matching signature with a real secret still returns `true`. `rg 'if \(!secret\)' packages/vercel/webhooks/types.ts`

msranjana:
hi @ambikeesshh want to work on this

PR #753, opened 2026-08-13
fix(intercom): fail closed on missing webhook secret

### LINKED ISSUE #711: fix(intercom): return error when webhook secret is missing
## What's wrong

`verifyIntercomWebhookSignature` in `packages/intercom/webhooks/types.ts` has no empty-secret guard. Compare is already `timingSafeEqual`.

## Change

Fail closed when `secret` is missing:

```ts
if (!secret) {
}
```

Leave the HMAC path alone.

## Check

Missing/empty secret returns that error. A correct signature with a real secret still passes. `rg -A3 'if \(!secret\)' packages/intercom/webhooks/types.ts`

MikeGatsby:
## Description

`verifyIntercomWebhookSignature` computed an HMAC signature against an empty
secret instead of rejecting the request outright when no secret was
configured. Every other webhook plugin in this repo (typeform, stripe,
notion, spotify, gitlab, mailchimp, and ~25 others) already fails closed on
a missing secret with the same `{ valid: false, error: 'Missing webhook
secret' }` shape — Intercom (and Vercel, tracked separately in #712) were
the two outliers.

Fixes #711

## Checklist

      (scoped to `@corsair-dev/intercom`)
- [x] I have run `pnpm test` and all tests pass (scoped to
      `@corsair-dev/intercom`; pre-existing unrelated failures in
      `api.test.ts` come from live Intercom API calls needing real
      credentials, confirmed present before this change via git-stash A/B)
- [x] I have added tests for the new guard
- [ ] Full-repo `pnpm lint`/`pnpm build` not run (no `lint` script exists
      at the package level; a full monorepo build was out of scope for a
      2-file fix — happy to run it if a maintainer wants it before merge)

## Additional Notes

No behavior change to the signature comparison path — only adds the
missing-secret guard before the HMAC is computed, matching the repo's
established pattern exactly.

### DIFF
--- a/packages/intercom/webhooks/types.test.ts
+++ b/packages/intercom/webhooks/types.test.ts
@@ -1,0 +1,69 @@
+import type { WebhookRequest } from 'corsair/core';
+import crypto from 'crypto';
+
+import type { IntercomWebhookPayload } from './types';
+import { verifyIntercomWebhookSignature } from './types';
+
+describe('verifyIntercomWebhookSignature', () => {
+	const secret = 'my-intercom-client-secret';
+	const payload: IntercomWebhookPayload = {
+		type: 'notification_event',
+		topic: 'ping',
+		id: null,
+		app_id: 'app123',
+		created_at: 1700000000,
+		first_sent_at: 1700000000,
+		data: {
+			type: 'notification_event_data',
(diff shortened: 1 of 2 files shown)


Dhirenderchoudhary:
CLosed as fixed in #817

PR #756, opened 2026-08-13
feat(googledrive): add storage.getQuota endpoint

### LINKED ISSUE #755: feat(googledrive): add storage.getQuota for Drive storage quota
### What problem does this solve?

Agents using the Google Drive plugin can list, upload, and delete files, but they cannot ask how much storage the connected account has left.

That is a real gap. Upload and copy flows have no way to fail closed on quota. A Workspace account with pooled or unlimited storage looks the same as a consumer account that is 1 byte from the cap. OneDrive already exposes this as `drive.getQuota`; Drive does not.

Google's surface is `GET /drive/v3/about`. Quota lives on `storageQuota` (`limit`, `usage`, `usageInDrive`, `usageInDriveTrash`). `about.get` also requires a `fields` mask — omit it and the call 400s. There is no open issue covering this.

### Proposed solution

Add `storage.getQuota` to `@corsair-dev/googledrive`.

- Group: `storage` (task-oriented, same pattern as the existing `search` group)
- Method: `getQuota` (matches OneDrive's `drive.getQuota`)
- Input: empty object
- Provider call: `GET /about?fields=storageQuota` — field mask is internal, not agent-supplied
- Output: `{ limit?, usage, usageInDrive, usageInDriveTrash }` — all int64-as-string, `limit` optional because Workspace pooled/unlimited accounts omit it
- Persistence: none. Quota has no stable entity id and the current schema only models files, folders, and shared drives
- Risk: `read`
- Tests: `api.test.ts` schema-parse against the live response + `integration.test.ts` asserting the `googledrive.storage.getQuota` event row
- Scope: `packages/googledrive/**` only (R1)

Docs: https://developers.google.com/workspace/drive/api/reference/rest/v3/about/get

### Alternative approaches

1. `about.getStorageQuota` — mirrors Google's `about` resource so later `importFormats` / `maxUploadSize` slot in without a rename. Rejected: `about` is a poor tool name for an LLM, and this plugin already prefers task-oriented groups (`search`).

2. `about.get({ fields })` — one endpoint for the whole About resource. Rejected: it leaks the mandatory field-mask footgun to the agent and returns a fat sparse object.

3. Persist quota in the plugin DB. Rejected: no stable entity id, values go stale immediately, and it would force a schema change for ephemeral data.

### Additional details

I want to implement this. Please assign the issue to me.

Happy to add a short demo recording of the live `about.get` call for R4 before human merge.

imohit159:
## Description

Agents using `@corsair-dev/googledrive` can now read Drive storage quota and usage via `storage.getQuota`. Before this, upload/copy flows had no way to fail closed on a full account, and Workspace unlimited accounts looked the same as a consumer account at the cap.

The call is `GET /about` with `fields=storageQuota` hardcoded internally so the agent cannot omit the required field mask and 400. `limit` is optional because pooled/unlimited Workspace accounts omit it. Quota is not persisted — it has no stable entity id.

Fixes #755.

## Checklist

Live `GET /drive/v3/about?fields=storageQuota` returning `200` and `storageQuota` (`usage`, `usageInDrive`, `usageInDriveTrash`, `limit`):

https://github.com/user-attachments/assets/499a9850-5482-4110-b23c-22919acee1b3

## Additional Notes

- Scope is `packages/googledrive/**` only (R1).
- Tests added in `api.test.ts` and `integration.test.ts`; both skip when credentials are missing, matching the rest of this plugin.
- Lint / typecheck / build / test boxes are unchecked because `node_modules` is not installed in this checkout. I will tick them after a local `pnpm install` + those four commands.
- No breaking change, no new dependencies, no schema/table change.

## Summary by CodeRabbit

* **New Features**
  * Added Google Drive storage quota support through a new `storage.getQuota` endpoint.
  * Reports total storage limits and usage, including Drive and trash usage.
  * Added endpoint schemas and public types for storage quota data.

* **Tests**
  * Added API and integration coverage to verify quota responses and event tracking.

### DIFF
--- a/packages/googledrive/api.test.ts
+++ b/packages/googledrive/api.test.ts
@@ -3,6 +3,7 @@ import { ApiError } from 'corsair/http';
 import { makeGoogleDriveRequest } from './client';
 import { GoogleDriveEndpointOutputSchemas } from './endpoints/types';
 import type {
+	About,
 	File,
 	FileList,
 	Permission,
@@ -637,6 +638,25 @@ describe('Google Drive API Type Tests', () => {
 			});
 
 			expect(true).toBe(true);
+		});
+	});
+
+	describe('storage', () => {
+		it('storageGetQuota returns correct type', async () => {
+			const response = await makeGoogleDriveRequest<About>(
+				'/about',
+				TEST_TOKEN,
+				{
+					method: 'GET',
+					query: {
+						fields: 'storageQuota',
+					},
+				},
+			);
+
+			GoogleDriveEndpointOutputSchemas.storageGetQuota.parse(
+				response.storageQuota,
+			);
 		});
 	});
 
--- a/packages/googledrive/endpoints/index.ts
+++ b/packages/googledrive/endpoints/index.ts
@@ -2,6 +2,7 @@ import * as Files from './files';
 import * as Folders from './folders';
 import * as Search from './search';
 import * as SharedDrives from './shared-drives';
+import * as Storage from './storage';
 
 export const FilesEndpoints = {
 	list: Files.list,
@@ -34,6 +35,10 @@ export const SharedDrivesEndpoints = {
 
 export const SearchEndpoints = {
 	filesAndFolders: Search.filesAndFolders,
+};
+
+export const StorageEndpoints = {
+	getQuota: Storage.getQuota,
 };
 
 export * from './types';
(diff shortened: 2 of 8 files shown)


ambikeesshh:
@greptileai

devjain32 (approved):

PR #757, opened 2026-08-14
fix(gitlab): use constant-time comparison for X-Gitlab-Token

### LINKED ISSUE #708: fix(gitlab): use constant-time comparison for X-Gitlab-Token
## What's wrong

GitLab already rejects a missing secret. The actual compare in `packages/gitlab/webhooks/types.ts` is still `token !== secret`.

## Change

Swap that compare for `crypto.timingSafeEqual` after a length check. Keep the existing error string (`X-Gitlab-Token does not match configured secret`).

Leave the missing-secret and missing-header paths alone. `packages/gitlab/webhooks/types.test.ts` already exists.

## Check

Matching token still valid. Mismatch and length mismatch still invalid. `rg 'token !== secret' packages/gitlab/webhooks/types.ts` should be empty.

yashksaini-coder:
## Description

`verifyGitlabWebhookSignature` (`packages/gitlab/webhooks/types.ts`) already fails closed on a
missing secret and a missing header, but the actual comparison was still `token !== secret`.
JS string equality short-circuits at the first differing byte, so response timing leaks the
configured webhook secret byte-by-byte to anyone who can POST to the webhook endpoint.

The compare HMACs both sides (mixer key, not an auth secret) and then
`crypto.timingSafeEqual` on the 32-byte digests, so timing cannot leak the secret
byte-by-byte _or_ its length. The existing error string
(`X-Gitlab-Token does not match configured secret`) is unchanged. Missing-secret
and missing-header paths are untouched.

Also unwraps a repeated `x-gitlab-token` header (`Array.isArray(raw) ? raw[0] : raw`) before
comparing. `WebhookRequest.headers` is typed `Record<string, string | string[] | undefined>`
(`packages/corsair/core/webhooks/index.ts:29`), so this is required to get a string, and it
matches the convention every other webhook verifier in the repo already uses.

No shared helper was extracted: `corsair/http` only exports HMAC-shaped verifiers
(`verifyHmacSignature`, `verifyHmacSignatureWithPrefix`, `verifySlackSignature`), none of which
fit a raw shared-secret compare.

Fixes #708

## Checklist

Two notes on the boxes above, so nothing is overstated:

**Lint.** `pnpm lint` reports 6 warnings, all pre-existing in `www/` and untouched by this PR;
`npx biome check` on the two changed files is clean.

**Tests.** In `packages/gitlab`, `api.test.ts` and `integration.test.ts` fail identically with
and without this change — they call the live GitLab API and need credentials CI does not have.
I verified this by stashing the change and re-running. The three suites that run offline all
pass.

**Docs.** No documentation change applies here — no public API, function signature, or error
string changed, so the existing docs stay accurate.

## Screenshots / Demos

https://gist.github.com/ambikeesshh/c9523b397dba62865b701c92bd4319fc

Recording of the verifier on this branch — `rg` from the issue, then the unit suite:

```
$ rg 'token !== secret' packages/gitlab/webhooks/types.ts
(no output)

$ npx jest webhooks/types.test.ts

PASS webhooks/types.test.ts
  verifyGitlabWebhookSignature
    ✓ should fail closed when secret is missing (2 ms)
    ✓ should fail closed when both secret and token header are missing
    ✓ should return invalid if the token header is missing (1 ms)
    ✓ should return invalid if the token does not match the secret
    ✓ should return invalid for a same-length token that differs
    ✓ should return valid when the token matches the configured secret
    ✓ should use the first value of a repeated token header
    ✓ should reject a repeated token header when the first value is wrong
    ✓ should handle a multi-byte secret without throwing

Tests:       9 passed, 9 total
```

Nine cases. Length mismatch and value mismatch share the same error; the new
repeated-header case checks we fail closed when the first value is wrong.

## Additional Notes

No new dependencies, no breaking changes — `node:crypto` is stdlib.

**Follow-up, out of scope here:** `packages/telegram/webhooks/types.ts:285` compares its secret
token with `===` and has the identical timing leak. R1 limits a PR to one plugin, so it needs
its own PR — happy to open one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved GitLab webhook signature verification for repeated headers and same-length mismatched tokens.
  - Added support for multi-byte secrets without errors.
  - Enhanced security by using constant-time, HMAC-based token comparisons.

### DIFF
--- a/packages/gitlab/webhooks/types.test.ts
+++ b/packages/gitlab/webhooks/types.test.ts
@@ -55,8 +55,6 @@ describe('verifyGitlabWebhookSignature', () => {
 	});
 
 	it('should return invalid for a same-length token that differs', () => {
-		// Exercises the timingSafeEqual compare itself — 'not-the-secret' above is
-		// short enough that it never gets past the length guard.
 		const sameLength = 'my-super-secret-toXen';
 		expect(sameLength).toHaveLength(secret.length);
 		const result = verifyGitlabWebhookSignature(
@@ -97,8 +95,6 @@ describe('verifyGitlabWebhookSignature', () => {
 	});
 
 	it('should handle a multi-byte secret without throwing', () => {
-		// 'sécret' is 6 chars but 7 bytes: a char-length guard would let this
-		// through to timingSafeEqual, which throws on unequal buffer lengths.
 		const multiByte = 'sécret';
 		expect(
 			verifyGitlabWebhookSignature(
--- a/packages/gitlab/webhooks/types.test.ts
+++ b/packages/gitlab/webhooks/types.test.ts
@@ -85,6 +85,17 @@ describe('verifyGitlabWebhookSignature', () => {
 		expect(result).toEqual({ valid: true });
 	});
 
+	it('should reject a repeated token header when the first value is wrong', () => {
+		const result = verifyGitlabWebhookSignature(
+			requestWith({ 'x-gitlab-token': ['wrong', secret] }),
+			secret,
+		);
+		expect(result).toEqual({
+			valid: false,
+			error: 'X-Gitlab-Token does not match configured secret',
+		});
+	});
+
 	it('should handle a multi-byte secret without throwing', () => {
 		// 'sécret' is 6 chars but 7 bytes: a char-length guard would let this
 		// through to timingSafeEqual, which throws on unequal buffer lengths.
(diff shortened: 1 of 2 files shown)


ambikeesshh:
@greptileai

ambikeesshh (commented):
lgtm now

devjain32 (approved):

PR #761, opened 2026-08-14
feat(bugsnag): add BugSnag Data Access API integration

### LINKED ISSUE #760: Integration request: BugSnag

BugSnag

### API documentation link

https://docs.bugsnag.com/api/data-access/

BugSnag is an application stability platform: when software crashes in production,
the crash lands here with its stack trace, the release it came from, the device it
happened on and how many users it has affected. Teams then triage it - assign it,
snooze it, resolve it, or discard it as noise.

Today that triage happens by a human reading a dashboard. I would like an agent to
be able to read and act on it directly, covering the 60 operations the catalog
lists:

- Structure: organizations (3), projects (4)
- Access: collaborators (11), teams (6)
- Stability data: errors (3), events (2), event fields (3), pivots (2)
- Shipping: releases (2), release groups, trends (1)
- Configuration: saved searches (4), integrations (6)
- Feature flags (2)
- Privacy: GDPR/CCPA event data requests and deletions (9)

The concrete case that motivated it: an agent can answer "which unhandled errors
appeared in release 1.5.0 and are affecting the most users" by reading errors and
events against the release list, and then act on the answer - bulk-resolving the
ones already fixed, or opening the right ticket through a configured integration -
without a person triaging a dashboard by hand.

_No response_

### Additional context

Base URL: https://api.bugsnag.com

Auth: Authorization: token <personal auth token>. Note this is the literal word
"token", not "Bearer". A personal auth token is created at Settings -> My Account
-> Personal Auth Tokens and is a single credential - there is no account id,
subdomain or second header to resolve. An X-Version: 2 header is accepted.

Rate limits: this is the useful part, and it differs from most APIs I have looked
at. BugSnag reports its budget on every successful response via x-ratelimit-limit
and x-ratelimit-remaining, so a client can pace itself proactively rather than only
reacting to a 429. The budget is per-endpoint rather than global - observed live:

  GET /user                          x-ratelimit-limit: 100
  GET /organizations/{id}/projects   x-ratelimit-limit: 100
  GET /projects/{id}/errors          x-ratelimit-limit: 30

Some sources quote a flat 500 requests per minute, which the headers contradict, so
the plugin reads the headers rather than hard-coding a figure.

Errors: two distinct envelopes, and both matter.

  {"errors":["release_stage_name can't be blank"]}   - 400, an array of strings
  {"status":404,"error":"Not Found"}                 - 404, a single string

The 404 shapes carry different meanings, which is worth knowing before treating a
404 as "no such record": {"status":404,"error":"Not Found"} means the route does
not exist, while {"errors":["Event data deletion not found"]} means the route
exists and the resource does not. I used that distinction to map the GDPR
endpoints, whose real path is event_data_deletions rather than data_deletions, and
which offer create and read-by-id but no collection GET.

Responses: some endpoints return 204 with no body at all - stability_trend is one -
so a client cannot assume every 2xx carries JSON.

List and single reads differ in shape. An event in a list carries 11 fields; the
same event read by id carries 20, adding threads, metaData, request, device, user,
breadcrumbs, feature_flags, correlation and session. A schema built only from a
list response would silently drop half the record.

Several endpoints need a parameter that is easy to miss: release_groups and
feature_flags require release_stage_name, and the trend endpoints require
buckets_count or resolution. Without them the response is a 400 rather than a
default.

Privacy: events carry end-user personal data. A notifier can attach a user block
with name and email, plus arbitrary metaData and request contents such as URLs and
headers. Collaborators carry names and email addresses. The plugin keeps all of
that out of the event log, recording identifiers and counts only.

Plan availability: two operations could not be reached on a free account - the
project network grouping ruleset and the saved search usage summary. Every path
tried returned the route-absent 404 envelope rather than a permission error, so
they appear to be enterprise-only. That will be stated plainly in the PR rather
than implemented from documentation alone.

Destructive surface: this API can delete an organization, delete a project, erase
every error in a project, run GDPR data deletions and regenerate a project's API
key - which would invalidate the key every deployed notifier uses. Those are
marked destructive and are never exercised by the live test.

Agam00:
## Description

Adds the BugSnag integration against the Data Access API.

All **60 catalog operations** are implemented, across 15 groups: organizations,
projects, collaborators, teams, errors, events, event fields, pivots, releases, saved
searches, trends, integrations, feature flags, and the GDPR event data requests and
deletions.

56 of the 60 are verified against a live account. The remaining 4 are deliberately
never called, and the reasons are in "Destructive surface" below rather than left
implicit.

Fixes #760

Docs: https://docs.bugsnag.com/api/data-access/
Catalog: https://corsair.dev/oss/bugsnag

## Operations

61 endpoints implement 60 catalog operations. The difference is one endpoint that is
**not** a catalog operation and is not counted as though it were: `projects.get`
(`GET /projects/{id}`). The catalog lists no get-single-project operation, but the
endpoint is real - 200 with all 32 fields - and every project-scoped operation needs
an id from somewhere. A test asserts 61 registered and 60 claimed, so the two numbers
cannot quietly become one.

| Group | Ops | Group | Ops |
| --- | --- | --- | --- |
| collaborators | 11 | teams | 6 |
| projects | 6 (5 catalog) | integrations | 6 |
| saved searches | 5 | GDPR deletions | 5 |
| GDPR requests | 4 | organizations | 3 |
| errors | 3 | event fields | 3 |
| events | 2 | pivots | 2 |
| releases | 2 | feature flags | 2 |
| trends | 1 | | |

Organizations are reached through `/user/organizations` rather than `/organizations`,
because the API answers it relative to whoever owns the token. There is no way to
list organizations globally.

## Recon was wrong about seven paths, and this is how they were found

Worth reporting rather than quietly fixing, because the failure mode generalises: on
this API a **wrong path can return a 404 that looks like a missing record**, which
sends you looking for the wrong thing.

| Operation | First mapped as | Actually |
| --- | --- | --- |
| Create / get / delete saved search | `projects/{id}/saved_searches/...` | top-level `/saved_searches` |
| Saved search usage summary | *"enterprise-only"* | `/saved_searches/{id}/usage_summary` |
| Network grouping ruleset | *"enterprise-only"* | `projects/{id}/network_endpoint_grouping` |
| Collaborator access details (x2) | `.../access_details` | `.../project_accesses` |
| Add team memberships (x2) | `.../memberships` | `.../team_memberships` |
| Feature flag summaries | `feature_flags/summaries` | `feature_flag_summaries` |

**Two operations had been written off as enterprise-only and are not.** Both answer
200 on a free account. The evidence for that conclusion had been a 404 - but against a
path that does not exist, which proves nothing about the plan. `feature_flags/summaries`
was the clearest tell: it answered `{"errors":["Must supply valid feature flag ID"]}`,
a complaint about a missing id on a path that has no id, because `summaries` was being
parsed as `{id}` by `feature_flags/{id}`.

The paths came from the API Blueprint at
`bugsnagapiv2.docs.apiary.io/api-description-document`, which is machine-readable but
marked deprecated - so every path from it was re-verified live rather than trusted. That
was the right call: the document contains at least one typo
(`/organzations/{organization_id}/teams`) and omits routes the API serves.

## Where the catalog and the API disagree

One place, and it is stated here because a reviewer checking against the catalog would
otherwise read the difference as a missing field.

**`CREATE_CUSTOM_EVENT_FIELD`** is documented as requiring `display_id` as a separate
input from `path`, with an example where the two differ:
`display_id: "custom.user.accountId"` alongside `path: "metaData.user.accountId"`.

The live API does not work that way. It derives `display_id` from `path` and ignores
whatever is sent. Four creates were tried - the catalog's own `custom.`-prefixed form, a
plain dotted name, a hyphenated name, and omitting the field entirely - and **all four
returned 201 with `display_id` equal to the `path` value**. Omitting it is accepted, so it
is not required either.

So `display_id` is deliberately absent from that operation's input. Accepting a field the
API discards would invite a caller to delete the field later by the id they chose and
receive a 404 - which is not hypothetical: it happened during recon and left a field on
the account.

Related, and worth knowing for anyone reviewing the cleanup logic: **a newly created
custom field does not appear in `GET /event_fields` immediately.** The create response
carries `reindex_in_progress`, and until that completes the list omits the field while
`DELETE` by its id still works. A cleanup that lists first therefore finds nothing and
reports success while the field survives - which is how the field above went unnoticed
across two sessions.

## The bulk-update vocabulary was enumerated from the API, not from prose

Worth calling out because it was **wrong in two directions at once**, and both the wrong
and the right list had exactly twelve entries - so no count check could have caught it:

- it invented `unassign`, which the API rejects by name;
- it omitted `snooze`, which is valid, and which the catalog description mentions.

Each of the twelve was confirmed by a PATCH naming the operation against a **non-existent
error id**, so nothing could be modified. `{"errors":["Operation is not included in the
list"]}` means the name is invalid; any other response means the name is accepted and
something else about the request is wrong. `unassign`, `unsnooze`, `reopen` and `archive`
were all rejected by name and are absent.

The same probe surfaced three conditional requirements that were not documented anywhere:
`snooze` requires `reopen_rules`, `link_issue` requires `issue_url`, and
`override_severity` requires `severity`. Those, plus `assign` needing an assignee of
either kind, are enforced by refinements on the input schema - because finding out after
the fact is worse here than elsewhere: the request applies to every id in the batch.

## Authentication

`Authorization: token <personal auth token>` - the literal word `token`, **not**
`Bearer`. This is the single easiest thing to get wrong on this API, so the client
sets it and a test asserts it.

Single credential: no account id, subdomain or second header to resolve, so no
resolution chain and no discovery fallback. `X-Version: 2` pins the Data Access API
version so a future default shift cannot silently change response shapes.

## Rate limiting

BugSnag is unusual and better than most here: it reports its budget on **every**
successful response, and the budget is **per-endpoint rather than global**. Observed
live:

```
GET /user                          x-ratelimit-limit: 100
GET /organizations/{id}/projects   x-ratelimit-limit: 100
GET /projects/{id}/errors          x-ratelimit-limit:  30
```

So a caller can pace proactively instead of only reacting to a 429. `readRateLimit`
exposes both headers for that purpose, and a live test asserts that two endpoints
report *different* budgets - which is what makes a single hard-coded figure wrong. Some
sources quote a flat 500 requests per minute, which the headers contradict, so nothing
here hard-codes a figure. The `Retry-After` retry path remains as the fallback.

## Query construction: the plugin builds its own query strings

This is the most substantive design decision in the PR, and it is a workaround for a
core limitation, so it is spelled out.

`getQueryString` (`packages/corsair/async-core/request.ts:62-76`) makes two choices
that are wrong for this API. BugSnag is a Rails API, and both were established by
probing rather than by reading:

**1. Arrays become repeated bare keys**, and the API rejects that form outright:

```
?collaborator_ids=<id>     -> 400 {"errors":["Collaborator_ids must be an array"]}
?collaborator_ids[]=<id>   -> 200 [{"collaborator_id":"...","project_count":1}]
```

**2. An array of objects loses the `[]` marker, and then fails silently.** For
`{filters: {'error.status': [a, b]}}` the serialiser emits
`filters[error.status][type]=...` twice; Rails resolves repeated identical keys
last-wins, so a two-comparison filter would quietly mean *only the second comparison*.
Not an error - a different query than the caller asked for, answered 200.

The form the API requires keeps each comparison's `type` and `value` **adjacent**, and
grouping them is rejected:

```
...[][type]=eq&...[][value]=open&...[][type]=eq&...[][value]=fixed   -> 200
...[][type]=eq&...[][type]=eq&...[][value]=open&...[][value]=fixed   -> 400
```

Pair-adjacency is therefore a requirement, not a preference, and a generic serialiser
cannot be relied on to preserve it. So `buildQuery` in `endpoints/shared.ts` assembles
the string explicitly for every operation - one mechanism rather than two - and the
tests assert the exact strings. A live test additionally sends both orderings and
asserts the grouped one is rejected, so this cannot silently regress into a
plausible-looking wrong query.

## Filters: an unrecognised field is silently ignored

Verified by effect, and worth flagging because it is a trap for any caller. On a project
whose three errors are `error`/`warning` severity:

```
filters[event.severity][][value]=info    -> 0 rows   (the real field)
filters[error.severity][][value]=info    -> 3 rows   (not a field - ignored)
filters[totally.made.up][][value]=zzz    -> 3 rows   (invented - ignored)
```

A mistyped filter field does not fail. It returns unfiltered data with a 200, so the
caller believes the filter applied. Valid names come from `eventFields.list` (39
built-ins on the recon project plus any custom fields), and that is documented on
`errors.list` rather than hidden in a helper.

Filters *are* genuinely applied - a valid non-matching value returns 0 rows. Establishing
that took two attempts: the first filtered on `open`, which every error already was, so
an unchanged count of 3 was the correct answer for both a working filter and an ignored
one. The test now filters on `fixed`.

## Pagination

There is **no envelope**. A list response is a bare JSON array, and the paging state
arrives in headers instead:

```
link: <https://api.bugsnag.com/projects/{id}/errors?base=...&offset=1&per_page=1>; rel="next"
x-total-count: 3
```

The shared transport cannot surface those: `request()` returns
`responseHeader ?? responseBody` (`packages/corsair/async-core/request.ts:379`), so
a call yields **either** the parsed body **or** one named header, never both. A plugin
therefore cannot return the rows and their `Link` header together, and a caller cannot
follow `rel="next"`.

These operations page instead by the same `offset` and `per_page` parameters the
`Link` URL itself uses. `per_page` is bounded at 100 client-side rather than by the API -
`per_page=1000` was answered 200, so the API enforces no ceiling and an unbounded value
would let one call pull an arbitrarily large page.

**Two limits a caller has to distinguish**, mapped live on the error list:

```
offset 0..2     -> 1 row each      (3 records exist)
offset 3..100   -> empty array     - past the end, the signal to stop
offset 1000+    -> 422, code 60000 - too deep to answer, a refusal
```

An empty page and a 422 mean different things, and only the first means "stop", so a
`isPaginationLimit` predicate separates that 422 from an ordinary validation failure and
the handler explains the depth limit rather than reporting a malformed request. The cap is
on `offset` alone, not `offset x per_page`: `per_page=100&offset=100` answers 200 while
`per_page=100&offset=9999` answers 422.

The 422 message suggests `sort=unsorted`, which is a dead end for this access pattern -
combined with an offset it answers `{"errors":["Pagination Offset is invalid"]}`. Offset
paging and unsorted results are mutually exclusive, so paging beyond roughly a thousand
rows needs the `base`/`Link` cursor the transport cannot surface. **That is a real
limitation of this plugin**, stated rather than left for a reviewer to find.

And offset is not honoured uniformly: the **project list ignores it outright** -
`offset=9999` returns the account's single project, and does not 422 either. So the strong
paging guarantees are asserted against the error list and the weaker true claim elsewhere.
Both of those corrections replaced earlier assertions that were simply wrong - one claimed
a high offset returns an empty page on the project list, and one claimed the same on the
error list.

## Errors

Two distinct envelopes, both observed live:

```json
{"errors":["release_stage_name can't be blank"]}   // 400, array of strings
{"status":404,"error":"Not Found"}                  // 404, single string
```

The two 404 shapes carry **different meanings**, and the handler branches on it - but
the rule is narrower than it first appears, and getting that wrong is what produced the
two false enterprise-only verdicts:

- A garbage **path parameter** still matches its route, so `projects/<garbage>` returns
  `{"errors":["Project not found"]}` - the *resource*-missing shape.
- Only a path matching **no route at all** returns `{"status":404,"error":"Not Found"}`.

So route-absent means the path is wrong, and it becomes evidence of a plan restriction
only once the plausible paths are exhausted. The live test asserts both **bodies**, not
just the status - an earlier version asserted only `status: 404` for both cases and so
proved nothing about the distinction it claimed to pin.

Several endpoints also require a parameter that is easy to omit and answer 400 rather
than applying a default: `release_stage_name` on release groups, `buckets_count` on
trends, `filters` on all four GDPR creates, `operation` on bulk update,
`filter_options` on a custom event field, and `collaborator_ids` on project access
counts. Each is required in the corresponding input schema, so the caller is told
locally instead of after a round-trip. `stability_trend` returns **204 with no body at
all**, so a client cannot assume every 2xx carries JSON.

## Persistence

4 entities mirrored: `organizations` (16 fields), `projects` (32), `collaborators` (18)
and `teams` (4). These are the structural side - which organizations exist, which
projects belong to them, who can see them, and how they are grouped. They change rarely
and are the lookup every other operation needs.

Everything else is returned but **not** mirrored, with a reason per family rather than
one blanket claim:

- **Errors and events** arrive continuously and are only meaningful against a time range
  and a filter. A local copy would mirror a firehose and be stale before it was read.
- **Trends and pivot values** are aggregates over a window, not records.
- **Releases and release groups** are an append-only history whose counters keep moving.
- **Saved searches** because `filters` can contain end-user identifiers - searching for
  one customer's email address is an ordinary support workflow.
- **Configured integrations** because they hold third-party credentials.
- **Event fields** are filter metadata rather than data.

Reads never evict, and a test sweeps every GET to prove it: a project or collaborator
dropping out of a list is usually a permissions change rather than a deletion, and the
mirrored row still resolves ids that older errors reference. An explicit delete evicts.

Eviction is **required** - raising `BugsnagMirrorEvictionError` rather than warning - for
collaborators and organizations, because those rows hold a person's name and email
address and an organization's `billing_emails`. Reporting the deletion as a plain success
while the row survives would tell the caller someone's data is gone when it is still
queryable. Projects and teams evict best-effort, since a stale row there is untidy rather
than a disclosure. `projects.regenerateApiKey` refreshes rather than evicts - the project
still exists, and the response carries its new key.

## Privacy

BugSnag carries three kinds of material that must not reach the event log, and the audit
payloads record identifiers, counts and field *names* only:

- **End-user personal data.** A notifier can attach a `user` block with a name and email
  address to every event, plus arbitrary `metaData` and `request` contents - URLs,
  headers, IP addresses. Confirmed live: a seeded event returned
  `"user": {"id": ..., "name": ..., "email": ...}`. A pivot on `user.email` returns a
  list of end-user addresses.
- **Collaborator identities** - names and email addresses of the people on the account.
- **Secrets.** An organization carries an `api_key`; a project carries `api_key` and
  `upload_api_key`; a configured integration carries a credential for another service.

Rather than assert this per operation - which only covers the operations someone thought
of - `endpoints.test.ts` runs a **sweep**: every one of the 61 operations is executed
against a response poisoned with a planted secret, email address, name and metadata
value, and none may appear in what is logged. A further test proves the poison really is
injected, so the sweep cannot pass vacuously.

Where a value is genuinely useful for an audit trail it is recorded deliberately: which
*fields* a filter used (never the values), whether `full_reports` was requested (because
that says whether personal data was pulled), and a GDPR deletion's `status` (because
"prepared" and "carried out" can be days apart).

## Destructive surface

This API can destroy an account, so the risk levels follow what an operation can destroy
rather than its HTTP method. 13 operations are `destructive`, and a test asserts the set
is exactly those 13 - so a new one has to be considered rather than added silently.

Two that would not be caught by a naming rule:

- `projects.regenerateApiKey` deletes nothing, but every deployed notifier stops
  reporting until it is redeployed with the new key.
- `errors.bulkUpdate` can apply `delete` or `discard` to an arbitrary batch, which is why
  `operation` is an enum rather than a free string.

And `dataRequests.*` creates are `write` rather than `read`: they destroy nothing, but
they gather everything the account holds about an identified person and return a download
link.

**How the destructive operations were verified.** Five were exercised live against a
**throwaway project** created for the purpose and deleted immediately - create project,
regenerate its API key, delete all its errors (it had none), configure an integration
(empty body, so validation rejected it), and delete the project. The account was
compared before and after and restored exactly, and the seeded errors on the real project
were confirmed untouched. The saved-search and team round trips were done the same way,
and a custom event field was created and removed.

**Four operations were never run at all**, and will not be:

| Operation | Why not |
| --- | --- |
| `organizations.delete` | Destroys the account |
| `collaborators.invite` | Emails a real person |
| `collaborators.updatePermissions` | Would alter the account's only admin |
| `collaborators.delete` | Would remove the account's only admin |

Their routes come from the blueprint and their shapes are covered by mocked unit tests.
**The live suite is read-only throughout** and will stay that way.

Provenance is tracked per operation rather than claimed in aggregate: 56 live, 4
deliberately not. Three response shapes could not be observed either - configured
integrations (nothing configured, and configuring one needs real third-party
credentials), feature flags (the account has none) and the GDPR request/deletion records
(creating one would export or destroy real data). Those schemas are `.loose()` with only
their key required, and `schema.test.ts` marks them `verified: false` rather than
asserting documented field names as though they had been seen.

## Checklist

- [x] One plugin per PR
- [x] Tests with real assertions
- [x] Description complete
- [x] Linked issue
- [x] Working proof in Screenshots / Demos - see that section
- [x] No secrets, no `eval`, no `new Function`
- [x] No `any` on exported surfaces
- [x] Registered in `packages/corsair/core/constants.ts`
- [x] `dist/` not committed

## Screenshots / Demos

image

## Verification

Run on Node 22.18.0. CI runs Node 24, so these are a proxy rather than proof.

| Check | Result |
| --- | --- |
| `pnpm lint` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm run validate:plugins` | SUCCESS |
| `pnpm run validate:docs` | SUCCESS |
| `pnpm build` (packages/bugsnag) | exit 0 |
| `npx jest --ci --testPathIgnorePatterns="api\.test\.ts\|integration\.test\.ts"` | 3 suites, 283 tests passed |
| `pnpm test:live` against a live account | 23 tests passed |

Test files: 4. Assertions: 265.

`integration.test.ts` is excluded by `testPathIgnorePatterns` in `jest.config.cjs` as
well as by the flag CI passes, and self-skips without `BUGSNAG_AUTH_TOKEN`, so a plain
`jest` in this package reaches no network at all. `pnpm test:live` runs it.

Beyond the suites, three checks were run against this diff:

- a **surface verifier** mapping every registered endpoint to a catalog operation id,
  reporting five numbers rather than one: 60 catalog, 60 mapped, 61 registered, 60
  claiming an id, 1 documented orphan, 0 remaining. Self-tested by planting an unknown
  endpoint, a claim on a non-catalog id, and two endpoints claiming the same id.
- a **secret scan** over every file in the diff, combining pattern matching with the
  actual credential and identity values handled during development. Self-tested against
  a directory of planted leaks.
- the **plugin PR gate**, simulated as a non-draft, since `gate.ts:68` short-circuits on
  drafts and a green check on a draft is a pass-through rather than a verdict.

## Scope

- new files under `packages/bugsnag/`
- `packages/corsair/core/constants.ts`, exactly +3/-0
- `pnpm-lock.yaml`

## Two core suggestions, deliberately not implemented

Both are out of scope for a plugin PR under R1, so they are noted here instead.

**1. `request()` returns body or one header, never both.**
`responseHeader ?? responseBody` (`packages/corsair/async-core/request.ts:379`) makes
header-driven pagination unreachable from a plugin: BugSnag publishes `Link` with
`rel="next"` and `x-total-count` on every list, and this plugin can surface neither, so
it pages by offset instead. The same limitation blocks exposing
`x-ratelimit-remaining` on the response of the call that reported it. An `ApiResult`-shaped
return, or an optional `responseHeaders` passthrough, would let plugins support both
without changing any existing caller.

**2. `getQueryString` cannot express a Rails-style array or array-of-objects.**
`packages/corsair/async-core/request.ts:62-76` emits repeated bare keys for arrays and
drops the `[]` marker for arrays of objects. The first is rejected by this API; the second
silently produces a different query. Any plugin talking to a Rails backend hits both. An
opt-in bracket mode - or a documented escape hatch for building the string - would save
each such plugin reimplementing it, as this one had to.

## Summary by CodeRabbit

* **New Features**
  * Added BugSnag integration with API-key authentication and 61 supported operations.
  * Added organization, project, collaborator, team, error, event, release, integration, feature flag, search, trend, and data privacy operations.
  * Added pagination, filtering, validation, caching, audit logging, and safe rate-limit retries.
  * Added consistent BugSnag entity and response handling.
* **Tests**
  * Added comprehensive mocked, schema, endpoint, and optional live integration coverage.


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
LGTM

Tested locally with API.

devjain32 (approved):

PR #768, opened 2026-08-14
feat(apininja): add API Ninjas schema and database entities with vali…

### LINKED ISSUE #765: Integration request: API Ninjas

API Ninjas

### API documentation link

https://api-ninjas.com/api

API Ninjas is a single API key that reaches about 150 unrelated data services -
weather, geocoding, stock prices, tax rates, dictionary lookups, validation,
image generation. For an agent, that is unusual and useful: most of the small
factual lookups an agent needs mid-task are behind this one credential instead
of behind fifteen separate signups.

The work I keep hitting is the "one fact away" problem. An agent is drafting an
email and needs to know whether tomorrow is a public holiday in Germany. It is
validating a signup and needs to know whether the address is a disposable
domain. It is summarising a support thread and needs a sentiment score. Each of
those is one call, and today each is a different integration.

### Scope

129 operations, grouped for review as:

| Group | Ops |
| --- | --- |
| Markets and securities | 18 |
| Entertainment and random | 18 |
| Location and places | 13 |
| Economics and tax | 13 |
| Text and language | 11 |
| Generators and utility | 10 |
| Reference and knowledge | 10 |
| Internet and domains | 9 |
| Transport and vehicles | 8 |
| Time and calendar | 7 |
| Validation and identity | 7 |
| Health and food | 5 |
| **Total** | **129** |

### Example workflow this enables

An agent processing an inbound lead:

1. validates the email address and checks it is not a disposable domain
2. looks up the IP the form was submitted from
3. resolves the company domain to a WHOIS record and a logo
4. checks whether today is a working day in the lead's country before promising
   a callback

Four operations, one credential, and today each one is a separate integration.

No webhooks and no Corsair triggers. API Ninjas is request/response only - it
has no event subscriptions, no callback registration, and nothing that pushes
to a receiver. Every one of the 129 operations is a synchronous call that
returns its answer in the response body.

### Additional context

All facts below were confirmed against live calls on 2026-08-14 rather than
read from the documentation, because on this API the two disagree in ways that
matter.

### Authentication

A single API key in an `X-Api-Key` header. No OAuth, no account-specific host,
no second credential, and the key never travels in the query string.

### Base URL

    https://api.api-ninjas.com/v1/<endpoint>

Flat and single-segment - there are no path parameters anywhere in the surface.
Every input is a query parameter. 127 of the 129 operations are `GET`; the two
NLP operations that take long text (text similarity, embeddings) are `POST`
with a JSON body.

### Rate limits

The free tier documents 3,000 calls per month and 100 per hour.

More importantly: **there are no rate-limit headers on a successful response.**
The header set on a 200 is CDN and gateway plumbing only - no
`RateLimit-Limit`, no `RateLimit-Remaining`, no equivalent. The client
therefore cannot throttle proactively and will react to 429 with exponential
backoff.

And quota exhaustion does not arrive as a 429. It arrives as:

    400 {"error": "Monthly quota exceeded. Consider upgrading your subscription..."}

which is the same status code as a validation error. The plugin's error handler
inspects the body text for that case so an exhausted quota is reported as a
rate-limit condition rather than as a caller mistake.

### Error codes

| Status | Body key | Cause |
| --- | --- | --- |
| 400 | `error` | invalid parameters, missing or invalid API key, premium-gated endpoint or parameter, quota exhausted |
| 404 | `message` | unknown endpoint |
| 502 | `message` | server fault - but also returned for a wrong parameter name |

Two different body keys, so an error extractor has to read both. There is no
401 and no 403 on this surface: a missing key returns 400, which means status
alone cannot separate an auth failure from a validation failure.

### Pagination

There is none. No limit/offset envelope, no cursor, no total count. Collection
endpoints return a bare JSON array capped server-side, and where a `limit`
parameter exists at all it is premium-gated.

### Free-tier behaviour, and why it shapes the schemas

This is the part that would be easy to get wrong. The free tier withholds data
in three different ways:

1. **Whole endpoint gated** - returns 400 with "This endpoint is available to
   premium subscribers only."
2. **One parameter gated** - the endpoint works, but only one way. Weather by
   `lat`/`lon` is free while `city` is premium; holidays for the current year
   are free while `year` is premium.
3. **Fields replaced inside a 200** - the response succeeds and a field that
   should hold a number holds a sentence instead:

       {"ticker": "AAPL", "name": "This field is for premium subscribers only.", "price": 306.04}

The third case is the dangerous one, and the wording is not consistent across
endpoints, so it cannot be reliably detected by matching a phrase. Output
schemas accept both the real type and a string for any field a paid plan might
mask, keep only the primary key required, and are declared `.loose()` - a
strictly typed number field would reject the whole row for a free-tier user.

Recon classification: 89 operations fully usable on the free tier, 22 returning
masked fields, 16 gated at the endpoint or parameter level, 2 whose route could
not be resolved at all.

### Risk levels and retries

128 of 129 operations are pure reads. Exactly one mutates state - the counter
endpoint, when called with `hit` or `value` - so it is the entire
non-idempotent set and is excluded from network-error retries.

### Persistence

Roughly a dozen stores of reference data are mirrored: airports, airlines,
aircraft, countries, cities and postal geography, animals, celestial bodies,
stock exchanges, S&P 500 constituents, universities, emoji and vehicle
specifications. On an API with a 3,000-call monthly budget the cache is not a
nicety - repeated lookups of stable reference data are what exhaust the quota.
Prices, random content, generators and user-supplied lookups are deliberately
not mirrored.

### Two operations I want to flag

Two catalog operations have no route I could get to respond:

- quote of the day - `quoteoftheday`, `quoteofday`, `quotesoftheday` and
  `dailyquote` all 404, and the API directory lists no such API, even though
  the fact, joke and trivia of-the-day endpoints all exist and answer 200.
- list stock tickers - `tickers`, `stocktickers` and `tickerlist` all 404, and
  the `ticker` endpoint requires a symbol.

If the catalog row carries the intended endpoint for either, I will implement
them; otherwise I would rather ship 127 and say why than guess a route into the
registry.

### Scope note

129 operations is larger than a typical plugin but they are flat and
independent - no shared envelope, no relationships, no pagination. I think it
reads fine as one PR. If maintainers would rather have it split, the natural
split is the twelve groups in Field 3; say which grouping you want and I will
follow it.

abhishek-2k23:
## Description

Adds an API Ninjas integration: 129 operations behind one API key, covering
weather and geocoding, markets and securities, economics and tax, text and
language, validation, transport, reference data, generators and entertainment.

API Ninjas is unusual for an agent integration in that it is not one product -
it is about 150 unrelated single-fact services behind a single credential. Most
of the small lookups an agent needs mid-task (is tomorrow a public holiday in
Germany, is this a disposable email domain, what is the sentiment of this
paragraph) are one call here instead of one signup each.

API documentation: https://api-ninjas.com/api

Fixes #765 

### Coverage

**129 operations across 12 groups**, matching the OSS catalog row. Every route,
method and parameter name was confirmed against live calls on 2026-08-15 rather
than transcribed from the documentation - which turned out to matter, twice:
see "Routes the documentation gets wrong" and "Parameters the documentation
over-states".

Risk levels: **128 read, 1 write** (`counter`, which increments a stored value).

| Group | Ops |
| --- | --- |
| Markets and securities | 18 |
| Entertainment and random | 18 |
| Location and places | 13 |
| Economics and tax | 13 |
| Text and language | 11 |
| Generators and utility | 10 |
| Reference and knowledge | 10 |
| Internet and domains | 9 |
| Transport and vehicles | 8 |
| Time and calendar | 7 |
| Validation and identity | 7 |
| Health and food | 5 |
| **Total** | **129** |

### Authentication

A single API key in an `X-Api-Key` header, declared as
`api_key: { account: ['one'] }` and read through `ctx.keys.get_api_key()`. No
OAuth, no account-specific host, no second credential. `keyBuilder` raises
`AuthMissingError` rather than sending an empty key, which the provider would
answer with "Missing API Key." - a confusing way to report a configuration gap.

The key never travels in a query string, so nothing here depends on
`SENSITIVE_QUERY_PARAMS`, and a test asserts that for all 129 operations.

### Three API versions behind one host

`https://api.api-ninjas.com/<version>/<endpoint>` - flat, single-segment, with
no path parameters anywhere in the surface, so there is no path interpolation to
get wrong. Every input is a query parameter.

The version is not uniform, and it is not cosmetic:

| Version | Ops | Notes |
| --- | --- | --- |
| v1 | 117 | the bulk of the surface |
| v2 | 11 | bin, earnings, holidays, income tax, interest rate, mortgage rate, quotes, random quotes, quote of the day, random user, random word |
| v3 | 1 | recipe, whose v3 shape returns structured ingredients where v1 returned a pipe-delimited string |

Two of those v2 endpoints answer `404` on v1, so a plugin that assumed one
prefix would ship them broken. The version travels with the endpoint rather than
being fixed in the client.

127 operations are `GET`; text similarity and embeddings are `POST` with a JSON
body.

### Routes the documentation gets wrong

Nine operations do not live at the endpoint their name implies, and this was
only discoverable by probing. The provider's own sitemap lists three routes that
answer `404`, and three live routes appear in neither the sitemap nor the API
directory:

| Operation | Documented or implied | Actually |
| --- | --- | --- |
| Generate barcode | `/barcode` (in sitemap) | `/barcodegenerate`, parameter `text` not `data` |
| Generate user agent | `/useragent` (in sitemap) | `/useragentgenerate` |
| Insider transactions | `/insidertrading` (in sitemap) | `/insidertransactions` |
| Generate sudoku | `/sudoku` | `/sudokugenerate` |
| Solve sudoku | `/sudoku`, POST | `/sudokusolve`, GET |
| Is public holiday | - | `/ispublicholiday`, undocumented |
| Is working day | - | `/isworkingday`, undocumented |
| Fact / joke / trivia of the day | - | `/factoftheday`, `/jokeoftheday`, `/triviaoftheday`, all undocumented |
| S&P 500 constituents | `/sp500constituents` | `/sp500` |

Two further operations - quote of the day and list stock tickers - had no
working route under any name I could guess, and turned out to be `/v2/quoteoftheday`
and `/v1/stockpricelist`. Two more endpoints, `/v1/sp500` and `/v1/airlines`,
have no documentation page at all: their parameters were confirmed one at a time
against live calls.

Parameter names were confirmed the same way, by sending a deliberately wrong
request and reading which parameter the error named: `postalcode` wants
`postal_code` (the wrong name returns a `502`), `county` wants `county` plus
`state`, `routingnumber` wants `routing_number`, `unitconversion` wants `amount`
plus `unit`.

### Parameters the documentation over-states

Every parameter the documentation marks required was probed by omitting exactly
that parameter and keeping the rest - 76 checks. Three are not enforced: the QR
code `format` (which defaults to PNG), and `make` and `trim` on the car endpoint
(any one filter is enough). Those three are optional in the input schema,
because a schema that demanded them would reject calls the API answers. The
other 73 are required, as documented.

`docs-contract.ts` records both the documented flag and the probed result, and
`schema.test.ts` asserts the schemas follow it.

### Free-tier behaviour, and what it does to the schemas

The free tier withholds data three different ways, and only one of them is
visible from the status code:

1. **The whole endpoint is gated** - 400, "This endpoint is available to premium
   subscribers only."
2. **One parameter is gated** - the endpoint works, but only one way. Weather by
   `lat`/`lon` is free while `city` is premium. Holidays for the current year
   are free while `year` is premium.
3. **Fields are replaced inside a 200** - the call succeeds and a field that
   should hold a number holds a sentence:

       {"ticker": "AAPL", "name": "This field is for premium subscribers only.", "price": 305.79}

The third case is the one that breaks a schema built from documentation. It
affects 22 operations, and the wording varies between endpoints - "This field is
for premium subscribers only.", "Only available for premium subscribers.",
"Available for premium subscribers only.", "premium subscription required.", and
a lowercase variant - so it cannot be caught by matching a single phrase.

Every field a paid plan might mask therefore accepts both the documented type
and a string. Everything else is `.nullable().optional()` and every object is
`.loose()` - a strictly typed number field would reject the entire row for a
free-tier user, and a rejected row is a lost row.

Classification across the 129: 89 fully usable on the free tier, 22 returning
masked fields, 16 gated at the endpoint or parameter level, plus two more that
mix types across rows (an income-tax bracket ceiling that is a number until the
top bracket sends text, a migration figure that is a number until it is not).

### Rate limiting and quota

**There are no rate-limit headers on a success.** The header set on a 200 is CDN
and API-gateway plumbing only - no `RateLimit-Limit`, no `RateLimit-Remaining`,
no equivalent - so the client cannot throttle proactively and reacts to 429 with
exponential backoff.

Quota exhaustion does not arrive as a 429:

    400 {"error": "Monthly quota exceeded. Consider upgrading your subscription..."}

That is the same status as a validation error, so the error handler inspects the
body and routes the quota case to a rate-limit error with `maxRetries: 0` - the
monthly allowance does not come back inside a retry window. Without that, an
exhausted quota looks like a caller bug for the rest of the month.

The free tier documents 3,000 calls a month and 100 an hour; 177 calls in one
hour during development were never throttled, so neither figure is hardcoded.

### Error handling

| Status | Body key | Cause |
| --- | --- | --- |
| 400 | `error` | invalid parameters, missing or invalid key, premium gating, quota exhausted |
| 404 | `message` | unknown endpoint |
| 502 | `message` | server fault - and also an unusable parameter |

Two different body keys, so the extractor reads both. There is no 401 and no 403
anywhere on this surface: a missing key returns 400 reading "Missing API Key.",
an invalid one returns 400 reading "Invalid API Key.". The handlers therefore
match on body text, in this order: quota, credentials, plan, unknown route,
then any other 400.

The 502-for-unusable-parameter behaviour is why 5xx is not blanket-retried: a
wrong postal-code parameter name and an unsolvable Sudoku both return one, and
retrying a malformed request five times only spends quota. A 500 or 503 gets two
attempts with backoff.

### Pagination

There is none - no limit/offset envelope, no cursor, no total count. Collections
return a bare JSON array capped server-side (airports at 10, emoji at 30), and
where a `limit` parameter exists it is usually premium-gated.

### Persistence

**13 entities mirrored**, all reference data: airports, airlines, aircraft,
road vehicles, countries, cities, universities, stock exchanges, S&P 500
constituents, emoji, animals, planets and stars.

The reasoning is quota rather than latency. On a 3,000-call monthly budget,
repeated lookups of data that does not change are what exhaust the key, so the
mirror is the difference between a plugin that lasts the month and one that
stops halfway through it.

Most of these endpoints return no identifier, so rows are keyed by their natural
key - an airport's ICAO code, a city's name and country - and the three vehicle
endpoints share one store behind a `kind` discriminator, because their natural
keys collide. Masked values are dropped rather than stored: "This field is for
premium subscribers only." is not a fleet size, and a cached row saying so would
outlive the plan that produced it.

Deliberately not mirrored, and asserted so by a test: every price (`stockprice`,
`cryptoprice`, `bitcoin`, `commodityprice`, `exchangerate`, `marketcap`,
`mortgagerate`, `interestrate`), everything random or daily, every generator,
and every caller-supplied lookup. Nothing in this API deletes, so there is no
delete to evict on; each row carries `captured_at` instead.

### Privacy

Most inputs here are impersonal, but several are not: email validation and the
disposable-domain check take an email address, phone validation takes a number,
IP lookup takes an address, and sentiment, similarity, embeddings, spell check,
profanity filtering, language detection, scraping and nutrition all take
arbitrary caller text.

Those inputs are reduced to a length before anything is written to
`corsair_events`, and the reduction is by field name rather than by call site,
so a new operation that takes text cannot opt out of it by accident.
`behaviour.test.ts` asserts that an email, a phone number, an IP address and a
sentence never appear in a logged payload.

### Retry safety

128 operations are pure reads. The single write is the counter endpoint called
with `hit` or `value`, and it is the entire non-idempotent set - derived from the
declared risk levels rather than from a name pattern, with a test asserting the
set is exactly `{ utility.counter }` so a future write cannot join it silently.

### Tests

**1,151 tests across 9 suites**, plus 8 live tests that skip without a key.
Coverage of the package source: 98% of statements, **99% of branches**.

| Suite | Covers |
| --- | --- |
| `routing.test.ts` | all 129 operations against a mocked transport: versioned URL, method, `X-Api-Key` header, key never in the query string, no `undefined` serialised into a query, and a coverage sweep asserting exercised equals registered |
| `schema.test.ts` | every documented parameter present, no undocumented parameter, requiredness matching the probed behaviour, every documented response field declared, and all 120 captured responses parsing through their own schemas |
| `behaviour.test.ts` | mirroring and its keys, masked values not stored, reads never evicting, a failing cache not failing the call, audit redaction, image format defaults and wrapping |
| `persist.test.ts` | every one of the 13 stores: key derivation and its fallbacks, unkeyable rows skipped, nested fields lifted, the three vehicle sources kept apart, a missing store tolerated, a failing write swallowed |
| `endpoints.test.ts` | registry invariants across the four parallel maps, risk levels, the non-idempotent set, key building, handler merging, and which handler claims each failure |
| `error-handlers.test.ts` | what each handler then does: quota never retried, 429 retried, `Retry-After` honoured, plan and auth failures separated from bad requests, 502 not retried, 500 retried twice |
| `client.test.ts` | version routing, query serialisation, credential placement, POST bodies, `Accept` override, 429 retry and 400 non-retry |
| `shared.test.ts` | the helpers that decide what reaches the cache and the log: placeholder detection, numeric coercion, key composition, audit redaction |
| `build.test.ts` | the built bundle: 129 operations still callable, schemas still validating, requests still versioned and authenticated, mirroring still working, `corsair` and `zod` still external, fixtures not shipped. Skipped when `dist/` is absent |
| `api.test.ts` | live: one endpoint per version, both methods, a premium rejection, an unknown route and a missing key. Skipped unless `APININJAS_API_KEY` is set |

The routing suite replays the response captured from the live API for each
operation, so handlers are exercised against payloads the provider actually
sent. `fixtures.ts` holds those captures; the one endpoint that returns
person-shaped values (the random user generator, whose people are invented by
the provider) has them replaced with obviously fictional ones.

Writing the helper tests found a real defect: the placeholder matcher keyed on
the bare word "premium", so a legitimate value like "Premium Economy" or a fund
named "... Premium Fund" would have been dropped from the mirror as though the
provider had withheld it. It now matches the subscription wording that all 38
observed placeholder variants share.

## Checklist

## Screenshots / Demos
image

### Verification

| Command | Result |
| --- | --- |
| `biome check packages/apininjas` | clean, 38 files |
| `pnpm typecheck` (`tsc --build`, whole repo) | 0 errors |
| `pnpm run validate:plugins` | passes for `apininjas` |
| `pnpm run validate:docs` | `[SUCCESS] Docs validation passed!` |
| `pnpm build` | build success, `dist/index.js` 164.97 KB |
| `jest` (package) | 1,151 passed, 9 suites |
| `jest api.test.ts` with a live key | 8 passed |

### Scope

38 files under `packages/apininjas/`, plus exactly +3/-0 in
alphabetically between `apilabz` and `apisports`. `pnpm-lock.yaml` also changes,
because the workspace gained a package. `dist/` is gitignored and not tracked.

### Known limitations

- **Nine operations could not be exercised end to end on the free tier**:
  world time, WHOIS, list stock tickers, earnings call transcript, currency
  conversion, exchange rate, inflation, interest rate and nutrition are gated at
  the endpoint or parameter level, or hit a per-endpoint quota. Their routes and
  parameters are confirmed, and their 400s are captured, but their success
  shapes are declared from the documentation rather than from a capture.
- **22 operations return premium placeholder prose** in place of values on a
  free key, so those fields accept both the documented type and a string. On a
  paid key the real types arrive and still parse.
- **The three image operations return an image, not JSON.** The shared transport
  decodes any non-JSON response as text, so vector formats survive exactly and
  raster formats do not. `qrcode` and `barcodegenerate` therefore default
  `format` to `svg` rather than to the provider's `png`, and report what they
  returned in `content_type`; `randomimage` is JPEG-only and its payload should
  be treated as opaque. See the core suggestion below.
- The `cars` endpoint is marked deprecated by the provider. It ships because the
  catalog lists it, and the deprecation is noted in its description.
- Documented rate limits (3,000/month, 100/hour) were not enforced during a
  177-call hour, so neither number is encoded in the client.

`getResponseBody` in `packages/corsair/async-core/request.ts` decodes any
non-JSON response with `response.text()`. That is lossless for SVG and EPS and
lossy for image bytes, so a plugin cannot return a PNG through the shared
transport - `packages/googledrive`'s `filesDownload` types its result as
`z.any()` for the same reason. A response mode that hands back an ArrayBuffer
(or a base64 string) would fix it for every provider with a binary endpoint.
Flagging rather than fixing, since this PR is confined to the plugin.

## Summary by CodeRabbit

* **New Features**
  * Added API Ninjas integration with 129 operations across location, calendar, internet, validation, markets, economics, text, utility, transport, health, reference, and entertainment services.
  * Added authenticated, versioned requests with retries, image responses, and categorized endpoint access.
  * Added optional caching for selected reference, location, transport, and market data.
  * Added sensitive-data masking and privacy-conscious audit logging.
* **Tests**
  * Added comprehensive integration, routing, schema, error-handling, persistence, build, and client coverage.


abhishek-2k23:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
Thanks for contribution 
Tested locally with API all good every test passed and LGTM

Dhirenderchoudhary (approved):

devjain32 (approved):

PR #776, opened 2026-08-15
feat(formbricks): add Formbricks integration plugin

### LINKED ISSUE #775: Integration request: Formbricks

Formbricks

### API documentation link

https://formbricks.com/docs/api-v2-reference

I want an agent to be able to run a Formbricks survey programme end to end without a
human opening the Formbricks UI: author the survey, publish it, watch responses land,
keep the contact list in sync with our own user records, and hand completions off to
downstream systems.

Concretely, the operations I need, by group:

- Surveys (4): list, create, update, delete. Update is what moves a survey from draft
  to inProgress, so it is the operation that actually publishes.
- Responses (4): list, create, update, delete.
- Contacts (8): list, get, create, bulk upload, update attributes, delete, plus the two
  ids the catalog still lists under the older "people" naming.
- Contact attribute keys (7): list, get, create, update, delete, plus the two ids the
  catalog still lists as "attribute classes".
- Contact attributes (1): list the stored values.
- Action classes (2): list, create - these are the triggers that decide when a survey
  is shown.
- Webhooks (5): list, get, create, update, delete.
- Teams (3): list, delete, and list the workspace-team assignments.
- Roles (1), account identity (3), health (2).
- Client API (5): record a display, create or identify a respondent, read the
  environment bundle, read a respondent's state.
- Storage (2): public and private upload grants.

The scenario driving it: an agent notices a support ticket pattern, creates a targeted
survey, publishes it, identifies the affected users as contacts so the survey reaches
exactly them, and then reads the responses back to decide what changed - all without a
person in the loop for the mechanical parts.

_No response_

### Additional context

Recon notes from building against a live Formbricks Cloud workspace.

AUTHENTICATION
A single personal API key, sent as an x-api-key header. Both OpenAPI documents declare
apiKeyAuth as the only scheme; there is no OAuth surface. No second credential and no
resolution chain.

Key scope matters more than the auth shape does. A workspace-scoped key and an
organization-scoped key reach different routes: GET /v1/management/me rejects an
organization-scoped key outright with a 400 that tells you to use /v2/me instead. Most
management routes need a workspace-scoped key.

TWO LIVE API VERSIONS
v1 and v2 are both current and neither is a superset. Contacts expose only GET on v1
and only POST on v2; attribute keys have no create or delete on v1. So the version is
chosen per operation by what each route actually serves, rather than picking one.

WORKSPACE ID IN THE BODY
Most writes require workspaceId in the request body, not just in the key's scope. It is
not universal - PUT /v1/management/responses/{id} accepts a body without it - so each
operation states its own answer. A caller finds their workspace id from /v2/me, whose
workspacePermissions array lists every workspace the key can reach.

Note the terminology: the catalog descriptions all say environmentId, but the live API
takes workspaceId. Formbricks renamed environments to workspaces. This affects every
operation's input, not only the routes that were renamed.

RESPONSE ENVELOPE
Every response on both versions is wrapped in { data }. v2 list endpoints add
meta: { total, limit, offset }. v1 lists carry no meta at all.

PAGINATION
Inconsistent per route, and worth flagging because the wrong parameter is accepted with
a 200 and silently discarded:

- v1 /management/surveys advances by offset.
- Everything else that pages advances by skip - responses, webhooks, teams,
  workspace-teams, and v2 contact attribute keys.
- Four v1 routes ignore limit entirely and return every row: contacts, action classes,
  contact attributes, and v1 contact attribute keys.

The meta envelope reports an offset field even on the routes that ignore the offset
parameter, so reading the envelope leads to the wrong conclusion. All of the above was
established by seeding rows and comparing which ids came back.

RATE LIMITS
None observed. The API sends no rate-limit headers of any kind - no x-ratelimit-*, no
retry-after - and 15 rapid consecutive calls all returned 200 with no 429. So there is
nothing for a client to read and back off against, which is worth knowing before
someone writes code expecting a header that never arrives.

ERROR STYLE
v2 returns a structured body: { error: { code, message, details: [{ field, issue }] } },
with 422 for validation failures and field-level detail in details. v1 is terser and
sometimes returns a bare message string. Two routes answer 500 where a 422 would be
expected: a bulk contact upload containing an unknown attribute key, and
PUT /v1/management/responses/{id} sent without a data field.

SURVEY STATUS AFFECTS THE CLIENT API
A survey in draft answers 403 "Survey is not accepting submissions" to a display. The
message reads like a permissions problem but is a status one - the survey has to be
moved to inProgress first.

CLIENT ENVIRONMENT CACHING
GET /client/{workspaceId}/environment is cached server-side with a 5 minute TTL, so a
survey or action class created moments earlier may not appear yet. The payload carries
its own expiresAt.

Agam00:
## Description

Adds the Formbricks integration: all 46 catalog operations, verified against a live
Formbricks Cloud workspace rather than transcribed from documentation.

Fixes #775

Formbricks is a survey platform. This plugin lets an agent author and publish a survey,
read responses back, keep contacts in sync, and manage webhook subscriptions.

**Operations: 46 catalog ids, 47 registered operations, 38 distinct routes.**

| Group | Ops | Notes |
| --- | --- | --- |
| Surveys | 4 | `update` is what publishes - it moves `draft` to `inProgress` |
| Responses | 4 | |
| Contacts | 8 | includes the two ids the catalog still lists as "people" |
| Contact attribute keys | 7 | includes the two listed as "attribute classes" |
| Contact attributes | 1 | values, read-only |
| Action classes | 2 | |
| Webhooks | 5 | managed as a resource - see triggers below |
| Teams | 3 | teams, plus workspace-team assignments |
| Account / roles / health | 6 | |
| Client API | 5 | displays, user identify, environment, respondent state |
| Storage | 2 | public and private upload grants |

**Why 47 operations for 46 ids.** Formbricks renamed "people" to "contacts" and
"attribute classes" to "contact attribute keys" and removed the old routes, but the
catalog still lists both names. `contacts.listPeople`, `contacts.getPerson`,
`contactAttributeKeys.listClasses` and `contactAttributeKeys.getClass` are **aliases
calling the same URLs** as their current-name counterparts, so no catalog id 404s for a
caller working from the older entries. They are not four extra capabilities. A test
asserts each alias still points at its primary's route, and that each emits its own
audit event so the two stay distinguishable in a log.

One operation, `contactAttributeKeys.update`, claims **no** catalog id - it edits an
attribute key's definition, which the catalog has no id for.

**Triggers: 0.** Formbricks does have outbound webhooks, and this plugin manages them
as a resource, but the OSS catalog lists no triggers so no Corsair webhook handlers are
registered.

**Auth.** One API key, sent as `x-api-key`. Key scope decides reach: an
organization-scoped key is rejected by `GET /v1/management/me` with a 400 pointing at
`/v2/me`. Most management routes need a workspace-scoped key.

**Two live API versions.** v1 and v2 are both current and neither is a superset -
contacts expose only GET on v1 and only POST on v2. Version is chosen per operation by
what each route actually serves.

**Pagination is per route, and getting it wrong is silent.** The wrong parameter is
accepted with a 200 and discarded, so this was established by seeding rows and comparing
returned ids, not by reading status codes:

- `v1/management/surveys` advances by `offset` - the only route in the API that does.
- Everything else pageable advances by `skip`.
- Four v1 routes ignore `limit` too and return every row, so those operations expose no
  paging parameters rather than advertising ones the API discards.

The `meta` envelope reports an `offset` field even on routes that ignore `offset`, which
is what makes reading the envelope misleading. Callers say `offset` everywhere; the
plugin translates to the wire name per route.

**Persistence: 5 entities mirrored** - surveys, action classes, webhooks, contact
attribute keys, teams. All configuration. Responses, contacts, contact attributes and
displays are **deliberately not mirrored**: they are collected from survey respondents.
An attribute *key* is configuration and is cached; an attribute *value* is somebody's
email address and is not.

**Privacy.** Respondent answers, contact attribute values and `userId`s are sent where
the API needs them and never logged - audits record key names and counts instead. A
webhook create returns a signing secret that no later read returns; it is passed to the
caller and stripped before mirroring. Storage returns S3 presigned POST fields carrying
`X-Amz-Signature`; same treatment. Tests assert each of these directly, including a
sweep that runs every operation against a response poisoned with respondent data.

**Schemas built from live responses**, captured 2026-08-15, with a test asserting every
captured field is declared. Only the primary key is required, because Formbricks omits
or nulls fields by plan and survey type and a rejected row is a lost row.

## Verification

```
lint / typecheck / validate:plugins / validate:docs   clean
build                                                 36.73 KB
unit tests                                            250 passed, 3 suites
live tests                                            27 passed, real workspace
catalog coverage                                      46/46 ids, 0 missing
```

The live suite writes and cleans up: each test creates what it needs, deletes it in a
`finally`, and the suite compares eight resource counts before and after - a leftover
fails the run. It is excluded from a default run and self-skips without a key.

## Scope

30 files in `packages/formbricks/`, plus `packages/corsair/core/constants.ts` at
exactly +3/-0 and `pnpm-lock.yaml`. Nothing else touched.

## Known limitations

Stated rather than hidden - all Formbricks-side, all verified by effect:

- **Several documented filters are accepted and ignored.** `GET_RESPONSES` takes
  `contactId`, `startDate`/`endDate`/`filterDateField` and `sortBy` with a 200 and
  applies none of them; `LIST_WEBHOOKS` rejects `surveyIds` as a string, accepts
  `surveyIds[]=`, then ignores it. **Not declared in the input schemas** - a `contactId`
  filter that silently returns every respondent's answers is worse than a missing one.
- **`GET /v1/management/contacts` cannot be paged at all** and returns every contact in
  the workspace. `GET v2/management/contacts` is a 405, so there is no alternative.
- **`UPDATE_CONTACT_ATTRIBUTES` has no management route** - five candidates answer 404
  or 405. Implemented over the client user route, so it is keyed by `userId` and creates
  the contact if that id is new. Marked `write` and non-idempotent for that reason.
- **Two upstream 500s**, both worked around: a bulk upload containing an unknown
  attribute key, and `PUT /v1/management/responses/{id}` sent without `data` (the schema
  requires `data` to turn that into a local validation error).
- The documented 207 partial-success on bulk upload was not reproducible - it answers
  200, or rejects the whole batch with 422.
- Self-hosted instances are supported through a `host` option but were not exercised;
  only Formbricks Cloud was.

The catalog descriptions are also environment-scoped throughout (`environmentId`) while
the live API takes `workspaceId`. That affects every operation's input, not only the
renamed routes.

## Checklist

Test run - 250 unit tests across 3 suites, plus 27 live tests against a real Formbricks
Cloud workspace:

image

## Additional Notes

No new dependencies. No changes outside the plugin beyond the three registration lines
in `constants.ts` and the lockfile.

Core changes I deliberately did **not** make, to keep the diff in scope: the pagination
translation and the presigned-upload handling both live in the plugin rather than in
core, even though other plugins may hit the same shapes. Happy to lift either into core
in a follow-up if you would prefer that.

## Summary by CodeRabbit

* **New Features**
  * Added Formbricks as a supported provider.
  * Added 47 operations covering surveys, responses, contacts, webhooks, teams, storage, account, client, and health management.
  * Added API-key authentication, configurable hosts, validation, pagination, caching, deletion handling, and rate-limit support.
  * Added privacy-conscious audit logging that excludes secrets and sensitive response data.

* **Tests**
  * Added extensive unit, schema, endpoint, transport, and optional live integration coverage.


Agam00:
Quick heads up before you review.

**The Formbricks docs do not match their actual API.** I found seven places where they disagree.
If you check my code against the docs, those seven will look like mistakes - they are not. For
example: the docs use one field name, the API wants a different one. The docs list filters that
the API accepts and then silently ignores. The docs point file uploads at a route that rejects
everything.

I tested every operation against a real Formbricks account instead of trusting the docs, so where
my code and the docs differ, the code follows what the API actually does. I can walk through any
of the seven if you want the detail.

**Two things you might want changed:**

1. There are 47 operations for the 46 in the catalog. Four are duplicates - the catalog still
   lists some things under old names Formbricks has since renamed. I added both names so nothing
   is missing, but I am happy to delete them if you would rather keep it lean.
2. Four list operations have no paging. Formbricks ignores paging on those routes and returns
   everything, so adding the option would have been misleading.

All 46 operations are implemented and working. 251 automated tests, plus 28 that run against a
live account and clean up after themselves.

ambikeesshh:
@greptileai

ambikeesshh (commented):
missing key now fails closed, live tests only delete what they create, and webhook urls are http/https only. code looks fine to me now

thanks!

ambikeesshh:
@greptileai review

ambikeesshh (commented):
looks fine to me
thanks!

devjain32 (approved):

PR #782, opened 2026-08-15
feat(altoviz): add altoviz pulgin integration

abhishek-2k23:
## Description

Adds an Altoviz integration: 67 operations behind one API key, covering
customers and customer families, suppliers, contacts, colleagues, products and
product families, sale invoices, credit notes, quotes, receipts, purchase
invoices, the accounting reference tables and webhook subscriptions.

Altoviz is a French invoicing and accounting platform. What makes it worth an
agent's time is that invoicing is where a lot of small-business admin lives and
almost all of it is mechanical: look up the customer, find the right VAT rate
for their region, draft the invoice, record the receipt when payment lands. The
irreversible step - finalizing a document, which is a legal act in French
accounting - is deliberately not in this PR, so an agent can prepare the work
and a human still signs it off.

API documentation: https://developer.altoviz.com

Fixes #TBD-ISSUE

### Coverage

**67 operations across 14 resource groups**, matching the OSS catalog row. Every
route, method, parameter name and response shape was checked against live calls
on 2026-08-15 - 331 of them, against a tenant seeded with real records so that
every operation had a subject - rather than transcribed from the provider's
OpenAPI document. All 67 returned a captured success. Everything created during
verification was deleted afterwards, and every collection was re-read to prove
it.

That mattered more than usual here: five of the behaviours the plugin is built
around appear in neither the OpenAPI document nor the catalog, and two of them
contradict the catalog outright. See "What the live API does not match".

Risk levels: **41 read, 15 write, 11 destructive**.

| Group | Ops |
| --- | --- |
| Customers | 8 |
| Sale credits | 7 |
| Sale invoices | 6 |
| Receipts | 6 |
| Account and reference data | 6 |
| Suppliers | 5 |
| Products | 5 |
| Customer families | 4 |
| Contacts | 4 |
| Colleagues | 4 |
| Product families | 4 |
| Sale quotes | 3 |
| Webhooks | 3 |
| Purchase invoices | 2 |
| **Total** | **67** |

### Authentication

A single API key in an `X-API-KEY` header, declared as `api_key: {}` and read
through `ctx.keys.get_api_key()`. No OAuth, no tenant subdomain, no second
credential. The key never travels in a query string, so nothing here depends on
`SENSITIVE_QUERY_PARAMS`, and a test asserts that for all 67 operations.

**A missing or invalid key returns 401 with a completely empty body** - zero
bytes, no `content-type`. Confirmed three ways: no header, an empty header, and
a well-formed key that does not exist. An error extractor that reads a body
would report an empty message for the most common misconfiguration there is, so
the 401 handler supplies its own text and `keyBuilder` raises `AuthMissingError`
rather than sending an empty header at all.

### Base URL

`https://api.altoviz.com/v1/<resource>` - one host, no tenant subdomain. Every
operation is under `/v1` except the health check, which is `/hello` with no
version segment.

### What the live API does not match

**`PUT` clears every field the body omits.** The catalog says of both the
customer and supplier updates: "Only fields that are provided will be updated;
omitted fields retain their current values." The opposite is true. A
`PUT /v1/customers/{id}` carrying `id`, `type` and `companyName`, against a
customer created with a full profile:

| Field | Before | After |
| --- | --- | --- |
| `email` | `corsair-recon@example.com` | `null` |
| `phone`, `cellPhone` | both set | `null` |
| `firstName`, `lastName`, `title` | all set | `null` |
| `internalNotes` | text | `null` |
| `billingAddress` | full address | city and zip `null` |
| `shippingAddress` | full address | `null` |
| `family` | the family | `null` |

Eleven fields destroyed to change one. **Every update operation in this plugin
is therefore read-modify-write**: it GETs the record, merges the caller's fields
over it and PUTs the whole thing, so a caller supplying one field gets the
catalog's documented behaviour rather than the provider's. The extra read is
noted in each update's description. `behaviour.test.ts` asserts, for all five
updates, that a single-field input produces a request body carrying every other
field unchanged.

**Nested references are matched by value, and `id` is `readOnly`.** This one is
asymmetric, which is what makes it dangerous:

| Sent | Result |
| --- | --- |
| `vat: { id: 67996 }` | 400, `La TVA n'existe pas.` |
| `family: { id: 1007 }` on a customer | **200, and the record comes back with `family: null`** |
| `vat: { rate: 20, region: "FR" }` | works |
| `unit: { code: "H" }` | works |
| `family: { label, number }` | works, resolves to the existing family |

The silent case is the one that ships as a bug: passing an id is the obvious
thing to do and it produces unattached records with a success status. Input
schemas take an id from the caller - which is what an agent has, and what the
mirror provides - and translate it into the value form before the call.

**An invoice line priced with `unitPrice` is worth nothing, and reports
success.** `SaleDocumentLine` has no `unitPrice` - the price field is
`taxExcludedPrice`. The spec declares `additionalProperties: false`, but the API
does not enforce it, so the field is not rejected. It is ignored:

| Line carries | Result |
| --- | --- |
| `unitPrice: 999`, no product | **200, and the invoice totals 0.00** |
| `unitPrice: 999` + `productId` | 200, priced from the product record instead |
| `taxExcludedPrice: 999` | 200, invoice totals 999.00 |
| any unknown key | 200, ignored, rest of the line prices normally |
| `vat: {id}` or `unit: {id}` on a line | 500 Internal error, nothing named |

So an agent sending the field name almost every other invoicing API uses gets a
**zero-value invoice and a success status**. The input schema names
`taxExcludedPrice`, rejects `unitPrice` with a message pointing at it, and
`schema.test.ts` asserts no line body can reach the transport carrying one.

**Numbering is a per-document-type precondition.** A create that needs the
server to allocate a number fails on a tenant whose sequence has never been
initialised - `La numerotation des Clients n'a pas ete initialisee` - and
initialising it is a UI action with no API route. Customers accept an explicit
`number` instead; quotes do not. The `number` field's description says so
rather than leaving a caller to discover it.

**Deleting a family refuses rather than cascading.** A customer family that still
holds a member returns `409` with a French message; once empty it deletes with a
200. So there is no cascade for eviction to mirror - but there is the opposite
problem: creating a customer, supplier or colleague **auto-creates a contact**,
and deleting the parent leaves that contact behind. The three parent deletes
evict orphaned contacts from the mirror for that reason.

**The catalog's Create Customer description documents values the API rejects.**
The catalog row reads "use type='Company' for business customers ... or
type='Individual' for personal customers". Both are refused:

    POST /v1/customers {"type":"Company"}    -> 400
    POST /v1/customers {"type":"Individual"} -> 400

The accepted enum is `Business | Consumer | Government`. A plugin written from
the catalog description would fail on its first write call. The schema uses the
real enum; the operation description says so explicitly, since callers reading
the catalog will otherwise supply the documented values.

**The spec's own `api-version` parameter breaks the health check.** `GET /hello`
answers 200 with the account identity. The same call carrying the documented
`api-version=v1` answers 400 with an empty body. `TEST_API_KEY` therefore takes
no parameters.

**The quote status filter is a generator artefact and does not work.** The
OpenAPI document emits `Status.From`, `Status.Status.From`,
`Status.Status.CustomerId` and so on for `GET /v1/salequotes`. Live,
`Status=Bogus` returns 200 - the filter is silently ignored - and
`Status.Status=Pending` returns 500. `LIST_SALE_QUOTES` ships without a status
filter rather than with one that does nothing. The invoice equivalent is real
and enforced: `Status=Bogus` there is a 400.

**`OrderBy` is accepted and ignored.** `OrderBy=bogusfield` returns 200 on every
list endpoint. It is exposed because the provider documents it, with the
behaviour noted in the description.

**The find routes return arrays, not objects.** The catalog describes
`FIND_CONTACT` as returning contact details and `FIND_PRODUCT` as returning
"the first matching product ... null if no product matches". All seven find
routes return a JSON array, empty when nothing matches. Output schemas are
arrays; "first match" is a client-side convenience, not a provider behaviour.

**Two catalog rows are the same endpoint.** `FIND_PRODUCT` and
`FIND_PRODUCT_BY_NUMBER_OR_ID` are both `GET /v1/products/find`, the second a
strict superset of the first. Both ship because both are in the catalog; the
issue asks whether maintainers would rather have one.

### Pagination

Eleven list endpoints share `PageIndex`, `PageSize`, `OrderBy` and `query`. The
response body is a bare JSON array - no envelope, no total, no cursor. Paging
state comes back in headers, and a shared helper reads all six:

    x-page-index   x-page-size   x-page-count
    x-record-count x-page-next   x-page-prev

`x-page-next` carries a relative URL whose path segment is capitalised
(`/v1/Customers?PageIndex=2&PageSize=1`) and does not match the lower-case route
that was called, so the helper reads its query string rather than requesting the
URL verbatim.

**`PageIndex` is 1-based**, which is the trap:

    GET /v1/customers?PageIndex=0
    400 {"errors":["'Page Index' must be greater than or equal to '1'."],"message":"Validation failed"}

Confirmed on all eleven. A client defaulting to zero - which most do - fails
every list call, so the input schema's minimum is 1 and a test asserts it.

### Error handling

Eleven distinct response shapes, all captured:

| Status | Shape | Cause |
| --- | --- | --- |
| 400 | `{"errors":[...],"message":"Validation failed"}` | parameter or body validation |
| 400 | `{"errors":[],"message":"<specific>"}` | a rule with its own message |
| 400 | `{"errors":null,"message":"<French>"}` | a business rule, e.g. `La TVA n'existe pas.` |
| 401 | empty body | missing or invalid key |
| 404 | `{"errors":[],"message":"<specific>"}` | known route, absent record |
| 404 | empty body | unknown route |
| 404 | RFC 9110 ProblemDetails | one route family answers `{"status","title","type"}` |
| 405 | empty body | wrong method on a real route |
| 409 | `{"errors":null,"message":"<French>"}` | delete refused, record still in use |
| 429 | plain text, no content-type | quota exhausted; carries `Retry-After` in seconds |
| 500 | `{"errors":[...],"message":"Internal error"}` or `{"errors":[],"message":"An error occured"}` | provider fault (their spelling) |

`errors` arrives as an array, an empty array, or `null` - three types for one
field - and three shapes carry no text at all, so the mapping from status to
Corsair error class cannot be driven by the body. The extractor reads `errors[]`,
then `message`, then ProblemDetails `title`, and falls back to a status-specific
sentence when the body is empty.

**Provider messages are not surfaced to callers.** The message language is
inconsistent - validation errors are English, business-rule errors are French,
on the same status codes - and some of them name .NET internals. They go to the
log; the error the caller sees is written by the plugin.

**Validation aborts the whole body, not the offending field.** One bad enum
value produces two errors: a .NET conversion failure naming an internal type,
and a spurious "The customer field is required." for a field the caller did
supply:

    {"errors":[
      "Error converting value \"Bogus\" to type 'System.Nullable`1[...CustomerType]'. Path 'type', ...",
      "The customer field is required."],
     "message":"Validation failed"}

Every enum in the surface is therefore validated by zod before the request goes
out - `CustomerType`, `ProductType`, `PaymentMethod`, `VatMode`, `VatRegion`,
`LineType`, `DiscountType`, `ClassificationType`, `InvoiceStatusFilter`,
`ReceiptLinkType` and `WebhookType` - so a caller gets a field-level message
instead of a .NET type name, and the second, misleading line is never surfaced
as a separate problem.

### Security

Two CodeQL alerts came back on this PR, both fixed:

- **Incomplete URL substring sanitization** (`routing.test.ts`). The fixture
  assertion checked `call.url.startsWith(BASE)` against
  `'https://api.altoviz.com'` (no trailing slash), which a host like
  `https://api.altoviz.com.evil.com` would also satisfy. It's a test
  assertion, not a runtime security control, but the check is now
  `new URL(call.url).origin === new URL(BASE).origin`, which compares the
  actual host rather than a string prefix.
- **Polynomial regular expression used on uncontrolled data.** `getUrl` in
  `packages/corsair/async-core/request.ts` resolves `{param}` placeholders
  with `/{(.*?)}/g`, which rescans to the end of the string from every
  unmatched `{` - a path like `{a{a{a{a...` with no closing brace is O(n²).
  That's shared core code outside this PR's footprint, so the fix lives in
  `makeAltovizRequest` (`client.ts`) instead: this plugin never uses the
  `{param}` substitution feature (every id is interpolated into the path
  before the call, and the one caller-supplied string, `internalId`, is
  `encodeURIComponent`-encoded first, which already strips raw braces), so a
  literal `{` or `}` reaching `makeAltovizRequest` can only mean something
  slipped through unencoded. It's now rejected there, before the path is
  handed to the shared transport, rather than patching the regex upstream.

### Rate limiting

**The limit is exactly 100 requests, and a success gives you no warning it is
coming.** The complete header set on a 200 is `content-type`, `content-length`,
`date` and `strict-transport-security` - no `RateLimit-Limit`, no
`RateLimit-Remaining` - so the remaining budget is not observable. But the 429
is real and precise. Measured twice, independently:

    429 after 100 successful calls in 24.6 s (4.1 req/s)
    body:    Too many requests. Please try again later.   (plain text, and no
                                                           content-type at all)
    headers: retry-after: 36

Both runs tripped at exactly 100 successes; `Retry-After` differed (13 s and
36 s) because it reports the time left in the current rolling window. Honouring
it returned an immediate 200 both times.

The client therefore **sleeps for `Retry-After` rather than doubling** - the
provider's number is authoritative and exponential backoff would simply waste
the difference. The 429 handler reads the body as text, because a
JSON-parsing path yields nothing for the one error a busy integration will
actually meet. Sustained bursting past that eventually fails at the connection
level rather than with a status, so network errors are handled separately from
HTTP errors.

This section was wrong in three earlier drafts of this PR: an early 30-call
concurrent burst passed cleanly and I concluded there was no limit. 30 is simply
under the quota. The limit surfaced during an end-of-session cleanup sweep, and
is now asserted by a test.

### Ids

Every path id in the surface is `int32` - the API answers
`400 "The value ... is not valid."` for a GUID or any other string. Input
schemas use integers. `internalId` is a separate, caller-supplied string used as
a query parameter on the find routes and interpolated into the path on
`/v1/customers/getbyinternalid/{internalid}`, where it is URL-encoded before
interpolation and a test asserts no `undefined` can reach a path.

### Persistence

**7 entities mirrored, all of them reference data:** units, VAT
rates, accounting classifications, customer families, product families,
products and customers.

The split is deliberate and a test enforces it. Units, VAT rates and
classifications are the accounting reference tables - they change when the tax
code changes, they are read on the way to every invoice line, and mirroring them
is what keeps a plugin that drafts invoices from re-fetching the French VAT
table on every call. Products and customers are catalog data: read far more
often than written, and both have delete operations to evict on.

The reference mirror earns its place twice: because nested references resolve by
value rather than by id, the mirror is also what lets a handler turn the id a
caller supplies into the `{rate, region}`, `{code}` or `{label, number}` form the
API requires.

**Invoices, credit notes, quotes and receipts are deliberately not mirrored.**
They are transactional financial records whose status changes server-side - a
draft is finalized, an invoice is marked paid - without the plugin being told,
and a cached invoice that says "draft" when the provider says "paid" is the kind
of wrong answer that costs someone money. Reads never evict; deletes evict via
`deleteByEntityId`.

### Retry safety

The non-idempotent set is **all 15 writes plus all 11 destructive operations**,
listed explicitly rather than derived from a name pattern, with a test asserting
the set equals exactly the non-read operations. A replayed
`POST /v1/saleinvoices` is a duplicate invoice, not a retry, and Corsair replays
the entire endpoint call on a network error
(`packages/corsair/core/endpoints/bind.ts:206`).

**`UNREGISTER_WEBHOOK` gets a guard.** `DELETE /v1/webhooks` takes `id` and
`url` and the spec marks **both optional**, so an empty call may remove every
webhook on the tenant. This was the one thing I refused to probe. The input
schema requires exactly one of the two, and a test asserts a call with neither
is rejected before it reaches the transport.

### Privacy

This is an accounting system, so nearly every input is personal, financial, or
both: names, emails, phones, billing and shipping addresses, company
registration details, line-item prices, receipt amounts and payment methods.

Audit payloads carry **operation names, entity ids and counts only**. No
amounts, no addresses, no names, no line items, no email addresses. The
allow-list in `endpoints/logging.ts` is deny-by-default - a parameter's value is
recorded only if it is explicitly listed, and the list admits only ids, enum
values and counts - so a parameter added by a future operation is protected
before anyone reviews it.

### Tests

**TBD-TEST-COUNT tests across TBD-SUITE-COUNT suites.**

| Suite | Covers |
| --- | --- |
| `routing.test.ts` | all 67 operations against a mocked transport: base URL, `X-API-KEY` header, key never in the query string, method matching risk level, no `undefined` interpolated into a path, `internalId` encoded, and a coverage sweep asserting exercised equals registered |
| `behaviour.test.ts` | read-modify-write on all five updates (a one-field input must not clear a tenth field), nested references translated to their value form, request body envelopes, `undefined` omitted, 1-based paging defaults, the six paging headers parsed and their absence tolerated, mirroring into the right store, eviction on delete including orphaned contacts |
| `endpoints.test.ts` | registry invariants, risk levels, the non-idempotent set, error-handler ordering, the `UNREGISTER_WEBHOOK` guard, audit-payload redaction |
| `schema.test.ts` | every captured field declared against `tools/altoviz-shapes.json`, key-only rows parse, every enum rejected client-side before the call, line bodies carrying no field the provider rejects, no transactional entity mirrored |
| `error-handlers.test.ts` | all eleven response shapes including the three empty-body statuses, `errors` as array/empty/null, 401 producing a message with no body to read from, 409 reported as still-in-use, 500 retried, 400 not |

Handler inputs are generated by walking each operation's own zod schema rather
than hand-written, so a schema change cannot leave a stale fixture behind, and
the match count is asserted before every loop so a loop over zero rows cannot
pass silently. Response fixtures are the real captures from the verification run,
with the fictional records that produced them; nothing in them is a real person,
company or document.

## Checklist

## Screenshots / Demos
image

### Verification

| Command | Result |
| --- | --- |
| `biome check packages/altoviz` | TBD-LINT |
| `pnpm typecheck` (`tsc --build`, whole repo) | TBD-TYPECHECK |
| `pnpm run validate:plugins` | TBD-VALIDATE-PLUGINS |
| `pnpm run validate:docs` | TBD-VALIDATE-DOCS |
| `pnpm build` | TBD-BUILD |
| `jest` (package) | TBD-JEST |

### Scope

TBD-FILE-COUNT files under `packages/altoviz/`, plus exactly +3/-0 in
alphabetically between `alphavantage` and `alttextai`. `pnpm-lock.yaml` also
changes, because the workspace gained a package. `dist/` is gitignored and not
tracked.

### Known limitations

- **The API describes 114 operations; the catalog lists 67.** The 47 outside the
  catalog include the direct counterparts of shipped operations - create
  supplier, create colleague, create/get/download sale quote, update product,
  update and delete contact, list products - plus 8 statistics endpoints, 13
  xlsx export endpoints and 3 product-image endpoints. I shipped the catalog
  list and asked in the issue rather than deciding unilaterally.
- **The document lifecycle is deliberately excluded.** `finalize`, `send`,
  `markaspaid` and `markasrefunded` all exist and none are in the catalog.
  Finalizing an invoice is a legal act in French accounting and is
  irreversible; an integration that can draft and delete drafts but not finalize
  seems like the right blast radius for an agent. Say if you disagree.
- **Quotes need their numbering sequence initialised in the UI before the API
  can create one at all.** On a tenant that has never used the module, every
  create returns `La numerotation des Devis n'a pas ete initialisee`, and unlike
  the customer sequence an explicit `number` does not bypass it. Once switched
  on, the three quote operations work normally - verified against a real quote
  (`DE001004`, created, found, listed, deleted, then 404). Worth knowing:
  **deleting a quote that does not exist returns 200**, not 404, so that
  operation cannot report a miss to a caller.
- **`CREATE_RECEIPT`'s `links` parameter cannot be reached through catalog
  operations.** A receipt can only be linked to a finalized document
  (`Impossible d'encaisser un document en brouillon ... vous devez le finaliser
  au prealable`), and finalize is not in the catalog. Receipts create fine
  standalone, which is what ships.
- **Three downloads return real PDF bytes** (`application/pdf`, 82 KB and 81 KB,
  with `content-disposition` naming the document number) and are affected by the
  core text-decoding limitation - see the suggestion below. The
  purchase-invoice download returns `application/pdf` despite the spec declaring
  `application/json`, and round-tripped an uploaded file byte for byte.
- **`UPLOAD_PURCHASE_INVOICE` is the only multipart operation in the surface**,
  and the only create with **no delete anywhere in the API** - not in the catalog
  and not in the OpenAPI document. An uploaded document can only be removed in
  the UI. It goes through `request` from `corsair/http` with a `FormData` body
  rather than a raw `fetch`.
- **`REGISTER_WEBHOOK` returns 201 with `id: 0`.** The real id appears only in
  `LIST_WEBHOOKS`, and that list is eventually consistent - a deleted webhook
  reappeared for one call about two seconds after its delete. The register
  handler therefore returns what the provider sent rather than inventing an id,
  and the description says to list for it.
- **`UPDATE_COLLEAGUE` rejects a partial body with a 500**, which
  read-modify-write incidentally solves.
- **Two spec endpoints outside the catalog answer 500** on a well-formed query
  (`/v1/colleagues/find`, `/v1/suppliers/find`). Neither is shipped. Mentioned
  because it is a signal about how much of the published document is generated
  rather than exercised.
- **The tenant used for recon is French**, so reference rows come back
  localised and accented. Schemas do not assume ASCII.

`getResponseBody` in `packages/corsair/async-core/request.ts` decodes any
non-JSON response with `response.text()`, which is lossy for binary. Three
operations here return PDF bytes (`saleinvoices/download`,
`salecredits/download`, `purchaseinvoices/download`) and the provider's export
routes return xlsx. `packages/googledrive`'s `filesDownload` types its result as
`z.any()` for the same reason, and the `apininjas` PR raised it for image bytes.
A response mode that hands back an ArrayBuffer, or a base64 string, would fix it
for every provider with a binary endpoint. Flagging rather than fixing, since
this PR is confined to the plugin.

## Summary by CodeRabbit

* **New Features**
  * Added Altoviz integration for customers, contacts, suppliers, products, families, invoices, credits, receipts, quotes, webhooks, and account data.
  * Added document upload/download, pagination, filtering, validation, API-key authentication, caching, and persisted data schemas.
  * Added Altoviz to the available provider list.
* **Bug Fixes**
  * Improved rate-limit retries, error handling, safe updates, reference resolution, and deletion cleanup.
* **Tests**
  * Added comprehensive coverage for endpoint behavior, routing, validation, caching, errors, pagination, and live API integration.


Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
@greptile check

devjain32 (approved):

PR #786, opened 2026-08-15
feat(habitica): add Habitica integration (70 ops)

### LINKED ISSUE #785: Integration request: Habitica

Habitica

### API documentation link

https://habitica.com/apidoc/

Habitica is an open-source habit tracker that models personal productivity as a
role-playing game: habits, dailies, todos and rewards are tasks, completing them
earns experience and gold, and parties and guilds turn the whole thing into
something done with other people.

That makes it an unusually good fit for an agent, because the entire loop is
mechanical. Deciding what to work on, recording that it happened, keeping a
shared challenge in order and reporting on what actually got done are all things
a person currently does by hand in a web app several times a day.

I would like an agent to be able to drive that directly, covering the 70
operations the catalog lists:

- Tasks: create, list, get, update, delete, score, reorder, checklist items,
  tagging, and the challenge-task operations (13)
- Tags: create, list, rename, delete (4)
- Challenges: create, read, clone, delete, join, leave, list by group, list for
  the user, CSV export (9)
- Groups: create, list, read (including the party and Tavern aliases), update,
  leave, members, invitations, quest invitations (11)
- Group chat: read, delete a message, mark seen (3)
- User: profile, updates, account reset, equipment, cards, pinned items, inbox
  messages, push devices, notifications (11)
- Authentication: local registration, local login, social auth (3)
- Outbound webhooks: create, list, enable (3)
- Reference data: content catalogue, status, world state, model paths, news,
  shops, coupon validation (10)
- Data export: user data, task history, inbox (3)

The concrete case that motivated it: an agent can answer "what did I actually
finish this week, and which of my dailies am I quietly failing" by reading tasks
and history, and then act on the answer - rescheduling a daily, retiring a habit
that never gets scored, or setting up a challenge for the part that needs other
people - without anybody clicking through the app.

_No response_

### Additional context

Base URL: https://habitica.com/api/v3 - the version is in the path. Three
operations sit OUTSIDE that base, at https://habitica.com/export/*.

Auth: two headers, x-api-user (the account's user id) and x-api-key (the API
token). Both are checked: a valid token paired with a different account's user
id is answered 401 "There is no account that uses those credentials", so the
user id is a second credential rather than a routing hint the token implies.
Unlike Harvest there is no discovery fallback available - every authenticated
route already requires the user id, so no route exists that could discover it.
Both values are self-serve from Settings -> Site Data.

A third header, x-client, is MANDATORY and is the detail most likely to trip up
a new client. Omitting it is a 400 "Missing x-client headers", not a 401, and
the requirement does not track authentication: /api/v3/content takes no
credentials at all and still rejects a request without it. Only /status
tolerates its absence. The documented UserID-AppName format is not enforced -
any non-empty string is accepted.

Errors: real HTTP status codes with a consistent envelope,
{"success":false,"error":"<code>","message":"<text>"}. Observed live: 400
BadRequest, 401 NotAuthorized / invalid_credentials, 404 NotFound, 429
TooManyRequests.

Rate limits: 30 requests per minute per user id, and the figure is exact - the
30th request in a burst was the one refused. This is the binding constraint on
the whole integration and it shapes how the live tests have to be paced. Two
details of the headers matter:

- x-ratelimit-reset is a JavaScript Date string ("Sat Aug 15 2026 16:43:00
  GMT+0000 (Coordinated Universal Time)"), not a number, so a client that
  parseInts it gets NaN.
- retry-after is fractional seconds ("21.069"), so an integer parse truncates
  and retries slightly early.

Pagination is inconsistent and worth knowing per route rather than in general.
GET /tasks/user takes no page or cursor at all and returns every task in one
response. GET /challenges/user REQUIRES page and answers 400 without it. GET
/groups/:groupId/members pages by a lastId cursor.

Three things in the catalog do not match the API, all confirmed live:

1. GET /content's `filter` parameter EXCLUDES the categories you name rather
   than selecting them. The catalog describes it as returning content "filtered
   by a specific category type"; asking for quests returns all 55 other
   categories and omits quests, with a 200 and nothing to indicate anything is
   wrong. The server's own helper calls the argument removedKeys. An
   unrecognised key is ignored silently. This one is easy to miss because the
   status code and the response size both look plausible.

2. DELETE_GROUP's documented fallback route does not exist. The catalog says the
   operation calls POST /groups/:groupId/leave and, only if that fails, DELETE
   /groups/:groupId. There is no DELETE route under /groups other than the
   chat-message one. Both a real route with a missing record and an unrouted
   path return 404, so the status settles nothing - the messages differ ("Group
   not found or you don't have access." versus "Not found.").

3. GET /models/:model/paths does not accept `task`, which the catalog lists as
   valid. The real vocabulary is user, tag, challenge, group, habit, daily,
   todo, reward - the four task types are addressed individually.

Also: the catalog gives the content catalogue as "~9MB"; it measured 2.65MB
across 56 top-level keys, well inside a 20 second transport timeout.

Non-JSON responses: four operations do not return JSON - /export/history.csv is
text/csv, /export/inbox.html is text/html, and the challenge CSV export is
text/csv on the versioned base.

The three /export/* routes are worth a note because the source is misleading.
They are routed through authWithSession rather than the authWithHeaders
middleware every /api/v3 route uses, which reads as though a browser session
were required. They work perfectly well with ordinary header authentication -
all three return 200.

Privacy: GET /export/userdata.json returns the account holder's email address
under auth.local.email, along with their full task and message history. Group
chat and the inbox are private correspondence, and group invitations are made by
email address or username. The plugin mirrors none of it and logs no message
text, no task text, no email address and no coupon code.

Unlike several recent integrations, Habitica generally VALIDATES enum parameters
rather than ignoring them - ?type=notAType on the task list is a 400. But that
is a per-route property, not an API-wide one: the /content filter ignores
unknown keys on the same API.

Agam00:
## Description

Adds a Habitica integration covering all 70 operations listed in the OSS
catalog: tasks and their checklists, tags, challenges, groups and group chat,
the user document and its notifications, local and social authentication, the
user's outbound webhooks, the static content catalogue and shops, and the three
whole-account data exports.

Habitica is an open-source habit tracker that models productivity as a
role-playing game. The whole loop is mechanical - decide what to work on, record
that it happened, keep a shared challenge in order, report on what actually got
done - which is currently a person clicking through a web app several times a
day. This lets an agent drive it directly.

Fixes #785

Docs: https://habitica.com/apidoc/
Catalog: https://corsair.dev/oss/habitica

## Operations

70 operations over 67 distinct routes. The difference is two documented
collapses, not missing work:

- `GET_CONTENT` and `GET_CONTENT_BY_TYPE` are one route separated by a query
  parameter.
- `GET_GROUP`, `GET_PARTY` and `GET_GROUPS_HABITRPG` are one route separated by
  the group id passed (a UUID, `party`, or `habitrpg`).

Each is registered separately so no catalog id 404s, each gets its own audit
event so a log stays readable, and the PR says plainly that they are not three
capabilities.

| Group | Ops | Notes |
| --- | --- | --- |
| Tasks | 13 | CRUD, score, reorder, checklist item update/delete, tagging, challenge tasks |
| Tags | 4 | create, list, rename, delete |
| Challenges | 9 | CRUD, clone, join, leave, list by group, list for user, CSV export |
| Groups | 11 | create, list, read (+2 aliases), update, leave, members, invite, remove, quest invite |
| Group chat | 3 | read, delete a message, mark seen |
| User | 11 | profile, update, reset, equip, cards, pinned items, inbox, push devices, notifications |
| Auth | 3 | local register, local login, social |
| Webhooks | 3 | create, list, enable |
| Content | 10 | catalogue, status, world state, model paths, news, shops, coupon validation |
| Exports | 3 | user data (JSON), history (CSV), inbox (HTML) |

The surface was verified operation by operation rather than in aggregate. A
script maps every catalog id to a method and path, resolves each against the
185 routes extracted from the Habitica server source, and then joins that to
what the code actually calls, so a route that is implemented, tested, and
pointed at the wrong endpoint cannot pass. Every check was self-tested against
planted faults first.

## Three places the catalog and the API disagree

All three were confirmed live and are pinned by tests in `integration.test.ts`,
so if Habitica ever changes them the tests fail rather than the comments
quietly becoming wrong.

**1. `GET /content?filter=` excludes what you name; it does not select it.**
The catalog describes the operation as returning content "filtered by a specific
category type". Asking for `quests` returns the other 55 categories and omits
quests, with a 200 and no indication anything is wrong:

```
no filter                  56 keys, 2713 KB
filter=quests              55 keys, 2491 KB   <- `quests` is the missing key
filter=notARealContentKey  56 keys, 2713 KB   <- unknown keys ignored silently
```

The server's own helper names the argument `removedKeys`. The behaviour is
passed through rather than inverted in the plugin, because reversing it here
would put this integration out of step with every other Habitica client and
would break silently if the API were ever fixed. It is documented where a caller
meets it, in the input schema and the endpoint.

**2. `DELETE_GROUP`'s documented fallback route does not exist.** The catalog
says the operation calls `POST /groups/:groupId/leave` and, only if that fails,
`DELETE /groups/:groupId`. There is no DELETE route under `/groups` other than
the chat-message one. Both a real route with a missing record and an unrouted
path answer 404, so the status settles nothing - the messages differ:

| request | message |
| --- | --- |
| `GET /groups/<ghost>` (real route) | `Group not found or you don't have access.` |
| `DELETE /groups/<ghost>` (under test) | `Not found.` |
| `DELETE /groups/<ghost>/nonsense` (control) | `Not found.` |

The fallback is therefore not implemented: it could only ever 404, and it would
make a failed legitimate leave look like a delete that also failed. A welcome
consequence is that the operation has exactly one effect, so replaying it after
a transport failure cannot leave *and* delete.

**3. `GET /models/:model/paths` rejects `task`,** which the catalog lists as
valid. The real vocabulary, enumerated by asking the API one value at a time, is
`user, tag, challenge, group, habit, daily, todo, reward` - the four task types
are addressed individually.

Also minor: the catalog gives the content catalogue as "~9MB". It measured
2.65 MB across 56 keys, comfortably inside the shared transport's 20 second
timeout.

## Auth and transport

Habitica uses **two** credential headers, `x-api-user` and `x-api-key`, and both
are checked - a valid token with another account's user id is answered 401
`There is no account that uses those credentials`. The user id is therefore a
second credential, declared as an account-scoped `user_id` key. Unlike Harvest
there is no discovery fallback to build: every authenticated route already needs
the user id, so no route exists that could discover it.

A third header, `x-client`, is **mandatory** and is the detail most likely to
trip up a new client. Omitting it is a 400 `Missing x-client headers`, not a
401, and the requirement does not track authentication - `/api/v3/content` needs
no credentials and still rejects a request without it. Only `/status` tolerates
its absence, so the plugin sends it on every request. It carries no user id, so
nothing account-specific reaches request logs.

Rate limiting is 30 requests per minute per user id, confirmed exactly by firing
until throttled - the 30th request was the one refused. Two header details
shaped the config:

- `x-ratelimit-reset` is a JavaScript `Date` string, not a number, so
  `parseInt` yields `NaN`. It is deliberately **left unconfigured** rather than
  named in `headerNames`, since configuring it would advertise pacing the plugin
  cannot do.
- `retry-after` is fractional seconds (`"21.069"`). The transport's integer
  parse truncates to 21, so the first retry fires slightly early and can draw
  one further 429 before the backoff spaces attempts out. Documented so that
  extra 429 is not mistaken for a defect.

Four operations do not return JSON - two exports are `text/csv` and `text/html`,
and the challenge export is CSV - so they use a separate transport. The three
`/export/*` documents also sit outside `/api/v3`. Worth noting because the
server source routes them through `authWithSession` rather than the
`authWithHeaders` middleware every other route uses, which reads as though a
browser session were required; all three answer 200 to ordinary header auth.

## Privacy

`GET /export/userdata.json` returns the account holder's email address under
`auth.local.email`, along with their whole task and message history. Group chat
and the inbox are private correspondence, and group invitations are made by
email address or username.

- Nothing personal is mirrored. The five mirrored entities are tasks, tags,
  challenges, groups and webhooks; the user document, inbox and chat are not.
- No message text, task text, email address, coupon code or push-device
  registration id reaches the event log. Audit payloads carry named identifiers
  and counts only.
- The three credential-minting operations record **that an attempt happened and
  nothing else** - not even the field names, because `fields: ["username",
  "password"]` sitting in a retained log is an invitation to widen it later.
- Export failures exclude the response body from the thrown error.
- Every fixture in the package is fictional. No captured response from the
  development account was reused as test data.

## Tests

149 unit tests across 4 suites, plus a 16-test live suite excluded from CI.

- `endpoints.test.ts` - every one of the 70 operations: the method and path it
  calls, what it mirrors, what it evicts, and exactly what reaches the event
  log. A coverage sweep asserts the operations exercised are precisely the
  operations registered, so one cannot be added without a test.
- `schema.test.ts` - every live-captured key is declared. This is not
  redundant with parsing: the entities are `.loose()`, so an undeclared key
  parses cleanly and only a name comparison catches the gap.
- `client.test.ts` - both credential halves sent, `x-client` on every request,
  the right base URL per route family, non-JSON handling, and rate-limit config.
- `error-handlers.test.ts` - every handler against **both** transports. The four
  non-JSON operations throw a plain `Error` rather than an `ApiError`, so a
  handler that only recognised the latter would silently stop retrying rate
  limits on exactly the slowest operations.
- `integration.test.ts` - live, self-skipping without credentials, and paced at
  2.6s because the 30-per-minute limit applies to the whole run.

The live suite earned its place: it caught two schema bugs the unit tests
structurally could not. `group.cron` is an object, not the boolean its name
suggests, and `group.archive` is a string, not an object. Both were written from
the field name and both were wrong; `.loose()` hides an undeclared key, but a
wrongly *typed* declared key only fails against a real payload.

The schemas were then checked against `GET /models/:model/paths` - Habitica's
own model definitions - which confirmed no declared field is absent from the
model and no declared type contradicts it. That comparison also supplied seven
fields the development account never exercised, with types taken from the model
rather than guessed.

## Checklist

## Screenshots / Demos
image

## Additional Notes

**Footprint.** `packages/habitica/` plus a three-line addition to
`packages/corsair/core/constants.ts`, no deletions. `pnpm generate:plugin` also
re-sorted three unrelated provider pairs (`agenty`/`agentql`,
`ambientweather`/`ambee`, `canvas`/`cal`/`canva`); those were reverted so the
registration diff is three insertions and nothing else.

**No webhooks.** The catalog lists 0 triggers. Habitica does have webhooks, but
they are the user's own outbound webhooks - Habitica calling a URL the account
holder nominates - not events delivered to Corsair. They are implemented as
ordinary operations, and the scaffold's `webhooks/` directory was deleted rather
than left as empty stubs implying a surface that does not exist.

**Two deliberate asymmetries, matching the catalog rather than the API.** The
API has four webhook routes and the catalog lists three (`subscribe` is an
update setting `enabled=true`). The API has create/update/delete for checklist
items and the catalog lists update and delete only. Neither gap is filled in,
because the catalog defines the surface and adding siblings would put this
plugin out of step with other consumers of it.

**Documentation** here means the TSDoc carried by the code: every non-obvious
decision is recorded where a reader meets it, with the evidence and date. This
plugin adds no `plugin-docs.yaml`, matching the 125 of 135 plugins that have
none.

**One catalog contradiction left unresolved on purpose.** `CREATE_GROUP` states
guilds were removed in August 2023 and only `party` works, while `GET_GROUPS`,
`GET_GROUP` and `DELETE_GROUP` all describe guild behaviour. Both cannot be
current. The plugin does not adjudicate: it sends what the caller asked for and
lets Habitica answer, because inventing a client-side restriction would break
callers if the catalog note is the stale half.

**`user.reset` was never exercised live.** It deletes every task on the account
and cannot be undone. It is marked `destructive`, and it clears the mirrored
task collection afterwards - the response is the reset user and names nothing it
removed, so without that the mirror would keep answering with the account's
entire previous task list.

## Summary by CodeRabbit

* **New Features**
  * Added Habitica as a supported provider.
  * Added integrations for authentication, tasks, tags, challenges, groups, chat, user accounts, webhooks, content, and data exports.
  * Added account-scoped API-key authentication and typed endpoint validation.
  * Added local synchronization for supported Habitica entities.
  * Added rate-limit handling with automatic retries and secure error reporting.
* **Tests**
  * Added comprehensive unit and integration coverage for requests, endpoints, schemas, errors, retries, and exports.

### DIFF
--- a/packages/corsair/core/constants.ts
+++ b/packages/corsair/core/constants.ts
@@ -77,6 +77,7 @@ export const BaseProviders = [
 	'googlemeet',
 	'googlesheets',
 	'grafana',
+	'habitica',
 	'hackernews',
 	'harvest',
 	'hashnode',
@@ -207,6 +208,7 @@ export const ProviderDisplayNames = {
 	googlemeet: 'Google Meet',
 	googlesheets: 'Google Sheets',
 	grafana: 'Grafana',
+	habitica: 'Habitica',
 	hackernews: 'Hacker News',
 	harvest: 'Harvest',
 	hashnode: 'Hashnode',
@@ -344,6 +346,7 @@ export type AllProviders =
 	| 'googlemeet'
 	| 'googlesheets'
 	| 'grafana'
+	| 'habitica'
 	| 'hackernews'
 	| 'harvest'
 	| 'hashnode'
--- a/packages/habitica/client.test.ts
+++ b/packages/habitica/client.test.ts
@@ -25,9 +25,32 @@ const API_TOKEN = '11111111-1111-4111-8111-111111111111';
 const CREDENTIALS = { userId: USER_ID, apiToken: API_TOKEN };
 
 let captured:
-	| { url: string; method: string; headers: Record<string, string> }
+	| {
+			url: string;
+			method: string;
+			headers: Record<string, string>;
+			body?: string;
+	  }
 	| undefined;
 
+// Every test here replaces the global fetch. Restoring it afterwards keeps this
+// file from deciding what any later suite sees.
+const realFetch = global.fetch;
+afterEach(() => {
+	global.fetch = realFetch;
+});
+
+/**
+ * Replaces `global.fetch` with a stub that records the request.
+ *
+ * The `as unknown as typeof global.fetch` cast at the end is deliberate and
+ * confined to this helper. A faithful `fetch` implementation would have to
+ * satisfy the whole `Response` interface - `blob`, `formData`, `clone`,
+ * `bodyUsed` and the rest - none of which the transport touches. Widening the
+ * stub to the real signature would mean writing a dozen unused members whose
+ * only effect is to obscure the four fields that matter: `ok`, `status`,
+ * `headers` and the body readers.
+ */
 function mockFetch(
 	payload: unknown,
 	{
@@ -50,7 +73,12 @@ function mockFetch(
 				headers[key.toLowerCase()] = value;
 			}
 		}
-		captured = { url: String(url), method: init?.method ?? 'GET', headers };
+		captured = {
+			url: String(url),
+			method: init?.method ?? 'GET',
+			headers,
+			body: typeof init?.body === 'string' ? init.body : undefined,
+		};
 		const body =
 			typeof payload === 'string' ? payload : JSON.stringify(payload);
 		return {
@@ -100,6 +128,34 @@ describe('Habitica transport', () => {
 
 			expect(captured?.headers['x-api-user']).toBeUndefined();
 			expect(captured?.headers['x-api-key']).toBeUndefined();
+		});
+
+		it('forwards the body on an anonymous POST', async () => {
+			// The three authentication routes are `authOptional` and POST their
+			// credentials through this helper. An earlier version destructured
+			// only `method` and `query`, so registration and login were sending
+			// an empty body - and asserting method and path alone did not notice.
+			mockFetch({ success: true, data: {} });
+			await makeHabiticaAnonymousRequest('user/auth/local/login', {
+				method: 'POST',
+				body: { username: 'someone', password: 'a-password' },
+			});
+
+			expect(captured?.body).toBeDefined();
+			expect(JSON.parse(captured?.body ?? '{}')).toEqual({
+				username: 'someone',
+				password: 'a-password',
+			});
+		});
+
+		it('sends no body on an anonymous GET', async () => {
+			mockFetch({ success: true, data: {} });
+			await makeHabiticaAnonymousRequest('content', {
+				method: 'GET',
+				body: { ignored: true },
+			});
+
+			expect(captured?.body).toBeUndefined();
 		});
 	});
 
@@ -219,6 +275,74 @@ describe('Habitica transport', () => {
 				}),
 			).rejects.toBeInstanceOf(HabiticaUserIdMissingError);
 			expect(captured).toBeUndefined();
+		});
+	});
+
+	describe('rate limiting on the raw-fetch paths', () => {
+		/** A 429 carrying Habitica's fractional Retry-After. */
+		function mock429(retryAfter = '21.069') {
+			global.fetch = (async () =>
+				({
+					ok: false,
+					status: 429,
+					statusText: 'Too Many Requests',
+					url: 'https://habitica.com/export/history.csv',
+					headers: new Headers({
+						'Content-Type': 'application/json',
+						'retry-after': retryAfter,
+					}),
+					json: async () => ({ error: 'TooManyRequests' }),
+					text: async () => '{}',
+				}) as unknown as Response) as unknown as typeof global.fetch;
+		}
+
+		it('preserves status and Retry-After through an export failure', async () => {
+			// These paths bypass the shared transport, so they get no ApiError.
+			// Throwing a bare Error would discard the delay and leave the handler
+			// retrying blind against a fixed one-minute window.
+			mock429();
+
+			await expect(
+				makeHabiticaExportRequest('history.csv', CREDENTIALS),
+			).rejects.toMatchObject({
+				name: 'HabiticaHttpError',
+				status: 429,
+				retryAfter: 21_069,
+			});
+		});
+
+		it('preserves them through a challenge CSV failure too', async () => {
+			mock429('5');
+
+			await expect(
+				makeHabiticaTextRequest('challenges/c1/export/csv', CREDENTIALS),
+			).rejects.toMatchObject({ status: 429, retryAfter: 5_000 });
+		});
+
+		it('keeps the fraction instead of truncating it', async () => {
+			// The shared transport parseInts to 21, retrying ~69ms early. Here the
+			// parse is ours, so it rounds up and never fires inside the window.
+			mock429('21.069');
+			await expect(
+				makeHabiticaExportRequest('history.csv', CREDENTIALS),
+			).rejects.toMatchObject({ retryAfter: 21_069 });
+			expect(21_069).toBeGreaterThan(21_000);
+		});
+
+		it('omits retryAfter when the server sent none', async () => {
+			mockFetch('nope', { status: 500, contentType: 'text/csv' });
+
+			await expect(
+				makeHabiticaExportRequest('history.csv', CREDENTIALS),
+			).rejects.toMatchObject({ status: 500, retryAfter: undefined });
+		});
+
+		it('ignores an unparseable Retry-After rather than passing NaN on', async () => {
+			mock429('not-a-number');
+
+			await expect(
+				makeHabiticaExportRequest('history.csv', CREDENTIALS),
+			).rejects.toMatchObject({ status: 429, retryAfter: undefined });
 		});
 	});
 
--- a/packages/habitica/client.test.ts
+++ b/packages/habitica/client.test.ts
@@ -1,0 +1,261 @@
+/**
+ * Transport-level guarantees.
+ *
+ * These are the invariants every endpoint inherits, so they are asserted once
+ * here rather than repeated 70 times: both credential halves are sent,
+ * `x-client` is always present, the two non-versioned base URLs are used where
+ * they should be, and a missing user id fails before a request is made.
+ *
+ * All credentials here are fictional.
+ */
+import {
+	HABITICA_API_BASE,
+	HABITICA_CLIENT_ID,
+	HABITICA_RATE_LIMIT_CONFIG,
+	HABITICA_ROOT_BASE,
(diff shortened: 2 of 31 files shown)


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary:
Aligned Habitica with the official API and live-tested the remaining ops.

- Schema matches official model fields (dropped the long comments)
- Task list uses `dueDate`, not `tagId`
- Challenge-task create returns one object  we normalize it to an array
- Member list sends official `limit`
- Public/Tavern chat is retired (400)
- Live tests cover the rest of the catalog routes

LGTM

devjain32 (approved):

PR #795, opened 2026-08-16
feat(circleci): add CircleCI integration

### LINKED ISSUE #794: Integration request: CircleCI

CircleCI

### API documentation link

https://circleci.com/docs/api/v2/

CircleCI is a continuous integration and delivery platform: every commit runs
through pipelines made of workflows made of jobs, and a team's whole delivery
picture - what shipped, what is flaky, what is slow, what is misconfigured -
lives behind its API rather than in any one dashboard view.

That makes it a good fit for an agent, because the questions people actually
ask about CI ("why did this pipeline fail," "which tests are flaky this
month," "is this env var still set on the right context," "which orbs is this
org allowed to pull from") require pulling together several endpoints by hand
today.

I would like an agent to be able to drive that directly, covering the 65
operations the catalog lists:

- Contexts (REST + GraphQL twins): create, read, list/store/remove env vars,
  restrictions, delete (11)
- Projects: create, read, delete (3)
- Project environment variables: create, list, delete (3)
- Groups: create, read, list, delete (4)
- Schedules: list (1)
- Pipelines: list, list for project, get config, trigger (4)
- Pipeline definitions: get, list (2)
- Workflows: list by pipeline, get summary, list jobs, list test metrics (4)
- Insights: flaky tests, project workflow summaries, org summary/plan
  metrics, branches, pages summary (6)
- Jobs (by number): details, artifacts, test metadata (3)
- Orbs and namespaces: details, versions, existence/id/source lookups, list,
  categories, validate config, namespace rename/delete/alias-delete (15)
- Organization: get (1)
- User: current user, info, collaborations (3)
- Runners: list (1)
- Usage export: create, get (2)
- Org URL orb allow-list: create, delete (2)

The concrete case that motivated it: an agent can answer "which tests have
been flaky on this project over the last month, and is the pipeline that
keeps catching them still using an orb version we meant to retire" by reading
insights and orb data together, and then act on the answer - updating a
schedule, tightening the org's orb allow-list, or rotating a context's
environment variable - without anyone clicking through four separate CircleCI
UI screens to piece it together.

_No response_

### Additional context

Architecture: this is the most structurally involved integration built so
far, because CircleCI's real operational surface spans FOUR separate
transports, not one:

- REST v2 (documented, circleci.com/api/v2) - 40 of the 65 operations
- GraphQL (graphql-unstable, introspection disabled) - 17 operations, used
  where a catalog description explicitly commits to it ("using the GraphQL
  API", "via the orbConfig GraphQL query")
- REST v3 (undocumented, strict JSON:API shape - {"data": [...], "page": {}})
  - 5 operations, found by reading circleci-cli's (CircleCI's own open-source
  CLI) Go source as ground truth, since v3 has no published spec
- Legacy v1.1 (flat JSON, project+build-number scoped rather than UUID
  scoped) - 3 operations, the only transport that resolves a job "by its
  number" the way three catalog operations require

Auth: single credential across all four transports - a personal API token,
sent as "Authorization: Bearer <token>" (the spec's own recommended scheme;
Circle-Token header, HTTP Basic and a deprecated query-string form all also
work). Every one of the spec's auth-scheme descriptions repeats the same
warning: "Project API tokens are not supported for API v2. Use a personal
API token" - confirmed live, since the wrong token type is easy to misread as
"the token is invalid" rather than "the token is the wrong kind."

GraphQL despite the CLI's own migration off it: circleci-cli's current source
calls REST v3 for orbs and namespaces now, not GraphQL. But the server still
answers real GraphQL queries with real data on the same token, and several
catalog operations commit explicitly to using it - so both are implemented
against their named transport rather than collapsed to whichever produces the
same effect.

The legacy v1.1 transport leaks personal data: GET_JOB_DETAILS's raw response
carries the triggering commit's real author email under
all_commit_details[].author_email. This plugin strips that field explicitly
before returning or logging anything - covered by a dedicated regression
test, not just a narrower declared schema, because every entity here is
.loose() and a .loose() schema does not strip an undeclared field at parse
time; only code that actually removes it does.

Rate limits: 300 requests per window (x-ratelimit-limit), confirmed from live
response headers. The window's length was not determined: x-ratelimit-reset
held steady at 1 across rapid successive calls rather than counting down,
which is the shape of a window-length field rather than a countdown-to-reset
one despite the header's name - so it is deliberately left unconfigured
rather than fed a value that would misrepresent it. retry-after is honoured
when CircleCI sends one. Separately, the usage-export operations are
documented by the catalog itself as limited to 10 per hour, independent of
the 300-request budget.

Masked secrets, checked live rather than assumed: a project environment
variable's value comes back as "xxxx" + the real last four characters; a
context environment variable's comes back as truncated_value, the last four
characters with no prefix. Neither is ever returned in full, not even
immediately after being set. Both masked forms are still treated as
sensitive - never logged, never mirrored - since a masked fragment is still
part of a secret.

Plan gating: Policy Management (9 routes) answers 403 with a structured body,
{"error": "...", "context": {"type": "insufficient_plan", ...}}, confirmed
live against the pricing page's own stated Scale-plan restriction. No
catalog operation uses Policy Management routes, so this never surfaces in
the shipped surface, but it is the kind of clean, structured plan-gate signal
worth knowing CircleCI provides.

Pagination: where CircleCI's v2 list routes paginate, the envelope is
consistently {"items": [...], "next_page_token": ...} (sometimes plus
total_count) - not a bare array, confirmed against both the spec and live
responses.

Agam00:
## Description

Adds a CircleCI integration covering all 65 operations listed in the OSS
catalog: contexts and their environment variables and restrictions (REST and
GraphQL twins), projects and project environment variables, org groups,
schedules, pipelines and pipeline definitions, workflows, insights, jobs (by
number), orbs and namespaces, the org URL orb allow-list, usage export, and
user/organization/self-hosted-runner reads.

CircleCI's own operational surface spans four separate transports - documented
REST v2, an undocumented REST v3, GraphQL, and a legacy v1.1 API - which this
plugin implements against each catalog operation's own named transport rather
than collapsing everything to whichever produces the same effect.

Fixes #794 

Docs: https://circleci.com/docs/api/v2/
Catalog: https://corsair.dev/oss/circleci

## Operations

65 operations across 18 resource families:

| Family | Ops | Notes |
| --- | --- | --- |
| Contexts (REST) | 6 | create, get, list/upsert env vars, create/delete restriction |
| Contexts (GraphQL) | 5 | create, delete, query, store/remove env var - separate catalog ids from the REST family above |
| Projects | 3 | create, get, delete |
| Project env vars | 3 | create, list, delete |
| Groups | 4 | create, get, list, delete |
| Schedules | 1 | list - the only schedule operation the catalog claims of the v2 spec's five |
| Pipelines | 4 | list, list for project, get config, trigger |
| Pipeline definitions | 2 | get, list |
| Workflows | 4 | list by pipeline, get summary, list jobs, list test metrics |
| Insights | 6 | flaky tests, project workflows, org summary, plan metrics, branches, pages summary |
| Jobs (by number, v1.1) | 3 | details, artifacts, test metadata |
| Orbs | 11 | details, version, id/existence/latest-version/source queries, list, categories, category-id query, namespace listing, validate config |
| Namespaces | 4 | query exists, delete, rename, delete alias |
| Organization | 1 | get (GraphQL) |
| User | 3 | current user, info, collaborations |
| Runners | 1 | list |
| Usage export | 2 | create, get |
| Org URL orb allow-list | 2 | create, delete |

Two catalog id pairs are documented aliases - two ids, one route - rather than
missing work:

- `LIST_INSIGHTS_SUMMARY` and `QUERY_PLAN_METRICS` both resolve to
  `GET /insights/{project-slug}/summary`; credit usage is one field inside
  the same summary object, not a separate concept.
- `GET_ORB_VERSION` and `QUERY_ORB_SOURCE` both send the identical GraphQL
  `orbVersion(orbVersionRef:)` query, because that field returns
  `{id, version, source}` as a unit and introspection is disabled, so there
  is no server-side way to split metadata from source.

Each alias is registered as two operations, each gets its own audit event so
a log stays readable, and each has a dedicated test asserting the routes (or
GraphQL query text) are byte-identical - so if CircleCI ever splits them, the
test catches it rather than the comment quietly going stale.

## The four-transport architecture

This is the structural fact that shapes the whole plugin, so it is worth
stating up front rather than discovering it file by file:

| Transport | Base | Shape | Ops | Why this one |
| --- | --- | --- | --- | --- |
| REST v2 | `circleci.com/api/v2` | JSON, `{"items": [...]}` list envelope | 40 | The documented, current spec |
| GraphQL | `circleci.com/graphql-unstable` | GraphQL, introspection disabled | 17 | Several catalog descriptions commit explicitly to it ("using the GraphQL API", "via the orbConfig GraphQL query") |
| REST v3 | `circleci.com/api/v3` | Strict JSON:API, `{"data": [...], "page": {}}` | 5 | Undocumented; found by reading `circleci-cli`'s own Go source, since no spec exists |
| Legacy v1.1 | `circleci.com/api/v1.1` | Flat JSON, project+build-number scoped | 3 | The only transport that resolves a job "by its number" per three catalog operations' own descriptions |

Two things worth naming plainly:

**GraphQL despite the CLI's own migration off it.** `circleci-cli`'s current
source calls REST v3 for orbs and namespaces, not GraphQL - CircleCI's own
tooling has moved on. But the server still answers real GraphQL queries with
real data on the same personal token, and the catalog's own descriptions
commit to it for 17 operations, so those are implemented against
`graphql-unstable` rather than silently redirected to whichever REST route
produces the same effect.

**The v1.1 job-detail response leaks the triggering commit's real email.**
`all_commit_details[].author_email` comes back in the raw response.
`jobs.getDetails` strips it explicitly before returning or logging anything -
by destructuring it out of the response object, not merely by declaring a
narrower output schema. Every entity in this plugin is `.loose()`, and a
`.loose()` schema does not strip an undeclared field at parse time; it passes
it straight through. This is asserted by a dedicated regression test.

Adding a second, third and fourth transport meant re-supplying, for each one,
everything the shared `request()` helper gives the first transport for free:
a typed error carrying the HTTP status, the parsed `Retry-After` value in
milliseconds, and a 20-second timeout. The REST v3 and legacy v1.1 transports
route through the same shared helper as v2, so they inherit all three
automatically. GraphQL goes through a raw `fetch` instead - the shared helper
treats a non-2xx status as the failure signal, but GraphQL reports failure as
a 200 carrying an `errors[]` array, which the helper cannot represent - so
its timeout, status-carrying error type, and `Retry-After` parsing are each
re-supplied by hand in `client.ts`, and `error-handlers.ts`'s rate-limit
handler is tested against both the REST and the GraphQL error shapes.

## Auth and rate limiting

Single credential across all four transports: a personal API token, sent as
`Authorization: Bearer <token>` - the spec's own recommended scheme (a
`Circle-Token` header, HTTP Basic, and a deprecated query-string form also
work). Every auth-scheme description in the spec repeats the same warning:
"Project API tokens are not supported for API v2. Use a personal API token" -
worth having verified live, since the wrong token type answers a plain 401
that reads like "invalid token" rather than "wrong kind of token."

Rate limiting is 300 requests per window (`x-ratelimit-limit`, confirmed live).
The window's **length** is deliberately left unconfigured:
`x-ratelimit-reset` held steady at `1` across rapid successive calls rather
than counting down, which is the shape of a window-length field, not a
countdown-to-reset field, despite the header's name - configuring it as a
reset countdown would misrepresent what it measures. `retry-after` is
honoured when CircleCI sends one, on every transport including GraphQL's raw
`fetch` path (see above). Separately, the catalog documents usage-export
creation as limited to 10 per hour, independent of the 300-request budget.

## Persistence

Seven entities mirrored: projects, contexts (with their env-var metadata and
restrictions), project environment variables, schedules, org groups, orb
URL-allowlist entries, and pipeline definitions. Only the primary key is
required on every entity; everything else is `.nullable().optional()`, and
every object is `.loose()`.

- Pipelines, workflows and jobs - transactional activity, appended
  continuously, meaningful only against a date range, the same reasoning
  Habitica applied to a task's history and Loyverse applied to receipts.
- Insights and usage-export data - derived reports over that same moving
  target.
- The orb registry (orbs, versions, namespaces, categories) - a shared public
  catalogue, not this account's data.
- Environment variable **values**, in either masked form - see Privacy below.

A few nested shapes (`config_source`, `checkout_source` on pipeline
definitions) are declared from the spec as opaque `.loose()` records rather
than fully typed, because populating them needs a GitHub App integration the
development account does not have - stated as such in the schema rather than
presented as captured.

Three delete operations mirror a required eviction - `contextsGraphQL.delete`,
`groups.delete`, `projects.delete` - meaning a local mirror-write failure
after a confirmed remote delete still raises, so the caller and the logs both
learn the mirror needs manual attention. Each of the three logs the deletion
**before** attempting that eviction, not after: the audit event asserts "the
remote record is gone," which is already true once the delete call returns,
and placing the log after a step that can throw would silently lose the audit
trail of a real, confirmed destructive action whenever the local write fails.
Covered by a dedicated, mutation-tested sweep across all three rather than
one instance.

## Privacy

- **Environment variable values are never mirrored, logged, or returned in
  full - on either transport, in either masked shape.** A project env var
  comes back as `"xxxx"` plus the real last four characters; a context env
  var comes back as `truncated_value`, the last four characters with no
  prefix. Neither is the plaintext, not even immediately after being set, but
  a masked fragment is still part of a secret, so neither field is treated as
  safe to echo.
- **The v1.1 commit-author email is stripped before the response leaves the
  endpoint**, not merely left undeclared - see above.
- **Orb YAML source is never logged.** `orbs.validateConfig`'s audit event
  records only the boolean `valid`, never the submitted YAML or the
  validation error text.
- Audit payloads carry named identifiers and counts only - swept across all
  60 `auditPayload(...)` call sites and every hand-written log payload; none
  logs a value, a YAML body, or an email address.
- No real credential, org id, project slug, or personal identifier from the
  development account appears anywhere in this diff. Every fixture is
  fictional.

## Tests

136 unit tests across 4 suites (99 `expect()` assertions by source count,
more at runtime through the table-driven cases), plus a 13-test live suite
excluded from CI.

- `endpoints.test.ts` (87 tests) - every one of the 65 operations: the
  transport, method and path (or GraphQL field) it calls, what it mirrors,
  what it evicts, and exactly what reaches the event log. A coverage sweep
  asserts the operations exercised are precisely the operations registered.
- `schema.test.ts` (15 tests) - every live-captured key is declared against
  every `.loose()` entity, plus primary-key-only parsing.
- `client.test.ts` (18 tests) - base URL and auth per transport, the v3
  envelope unwrap, array query-param serialisation, GraphQL failure shape,
  and `Retry-After` parsing on the raw-fetch GraphQL path specifically.
- `error-handlers.test.ts` (16 tests) - every handler, including the 403
  "Permission denied" ambiguity CircleCI's own context routes have (it
  answers the same way for "no access" and "does not exist"), and the
  `Retry-After`-to-`headersRetryAfterMs` passthrough on both error types.
- `integration.test.ts` (13 tests) - live, self-skipping without credentials,
  paced at one request per 2.5 seconds, covering all four transports with
  real requests including two create-then-delete probes cleaned up in
  `finally`.

The live suite caught a real defect during this build: four v2 list
operations (`projectEnvVars.list`, `schedules.list`, `contexts.listEnvVars`,
`groups.list`) were treating CircleCI's `{"items": [...]}` list envelope as a
bare array. The mocked unit tests had not caught it because their fixtures
independently encoded the same wrong assumption - a live call against the
real API is what surfaced it, and both the code and the fixtures were fixed
together.

## Checklist

## Screenshots / Demos
image

## Additional Notes

**Footprint.** `packages/circleci/` (34 TypeScript files) plus a three-line
addition to `packages/corsair/core/constants.ts`, no deletions.

**No webhooks.** The catalog lists 0 triggers. CircleCI does have its own
outbound webhook resource (create/list/update/delete), but no catalog
operation manages it, so this plugin does not implement it at all rather
than partially cover a surface nobody asked for. `CIRCLECI_TRIGGER_PIPELINE`
is unrelated - it starts a pipeline run, not a webhook subscription.

**Operations confirmed live vs. mapped from source and spec.** 41 of the 65
operations were confirmed against the real API during recon and build; the
live regression suite added in this PR re-confirms a cross-transport sample
of those end-to-end on every run, including `pipelines.listForProject`,
which recon alone had left mapped but unfired. The remaining operations are
mapped from `circleci-cli`'s source and the OpenAPI spec but were not fired
live, mostly because they are destructive on the only real project this
account follows (`projects.delete`, `groups.delete`, `namespace.delete`),
rate-limited independent of the request budget (`usageExport.create`, 10/hour
per the catalog), or need account state this development account does not
have (pipeline definitions need a GitHub App integration; org group creation
answered a real, confirmed 403 whose cause - a plan restriction or a
personal-GitHub-account limitation - was not narrowed further, so the
create operation is documented as unverified-write rather than silently
dropped).

**Orb allow-list creation returns a partial shape on purpose, not by
omission.** `POST` to create an entry returns only `{id, message}`, not the
full record - confirmed live by creating and immediately deleting a real
entry. A caller wanting the full record needs a follow-up `GET`. The output
schema matches what the API actually sends back rather than the fuller shape
the request body implies.

`corsair/http` or any other file outside `packages/circleci/` - the array
query-param serialisation, the v3 JSON:API envelope, and the GraphQL error
shape were all handled entirely within the plugin.

## Summary by CodeRabbit

* **New Features**
  * Added CircleCI as a supported provider.
  * Added access to projects, pipelines, workflows, jobs, contexts, groups, namespaces, orbs, runners, schedules, usage exports, insights, users, organization data, and pipeline definitions.
  * Added REST, GraphQL, and legacy API support with pagination.
  * Added local caching, audit logging, and secure handling of sensitive values.
* **Reliability**
  * Added rate-limit retries, timeouts, authentication and permission handling, not-found detection, and clearer API errors.


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile review

Dhirenderchoudhary (approved):
Context is just `id`, `name`, `created_at`. Groups lost the fake `org_id`. Restrictions got `project_id`. Env vars and collaborations pick up the keys the spec actually has.

LGTM

devjain32 (approved):

PR #797, opened 2026-08-16
feat(doppler): add Doppler integration

### LINKED ISSUE #796: Integration request: Doppler

Doppler

### API documentation link

https://docs.doppler.com/reference/api

Doppler is a secrets manager: projects hold environments (dev/staging/prod),
each environment holds branch configs, and each config holds the actual
secret values a team's services read at runtime. The operational questions
people ask about it - "what changed in this config last week," "which service
tokens still have write access," "is this secret documented" - span several
resource families that today only come together by hand in the dashboard.

I would like an agent to be able to drive that directly, covering the 62
operations the catalog lists:

- Workplace: get, update (2)
- Workplace users: list, get (2)
- Workplace roles: list, get, list permissions (3)
- Activity logs: list, retrieve (2)
- Projects: list, create, get, update, delete (5)
- Project roles: list, get, list permissions (3)
- Project members: list, get, delete (3)
- Environments: list, create, get, delete, rename (5)
- Configs: list, create, get, update, delete, clone, lock, unlock (8)
- Config logs: list, get, rollback (3)
- Secrets: list, get, delete, update, download, list names, update note
  (project-scoped and config-scoped, two distinct routes) (8)
- Dynamic secrets: revoke lease (1)
- Service tokens: list, create, delete (3)
- Integrations: list (1)
- Invites: list (1)
- Groups: remove member (1)
- Webhooks: add, list, get, update, delete, enable, disable (7)
- Change requests: list (1)
- Share: create plain-text link, create end-to-end-encrypted link (2)
- Auth: get authenticated user info (1)

The concrete case that motivated it: an agent can answer "which service
tokens on this project can still write secrets, and does the STRIPE key in
prod have a note explaining what it rotates to" by reading service tokens and
a secret's note together, and then act on the answer - revoking a stale
token, locking a config before a release, or generating a time-limited Share
link for a teammate who needs one credential without dashboard access -
without anyone clicking through several separate Doppler screens to piece it
together.

_No response_

### Additional context

Source of truth: Doppler publishes no single downloadable OpenAPI spec.
DopplerHQ/cli (Go, Apache-2.0) covered 37 of the 62 operations via its own
Go source. The remaining 25 were resolved by checking the docs site's
robots.txt, finding a linked llms.txt machine-readable index, and discovering
that each individual reference page (docs.doppler.com/reference/<slug>.md)
embeds a real per-operation OpenAPI 3.1 fragment in a fenced code block -
exact method, path, and full request/response schema for every one of the 62
operations, cross-checked against the CLI source and live calls.

Auth: a single Bearer token authenticates both transports this plugin uses -
the documented REST v3 API and Doppler Share's /v1/share routes. Confirmed
live: Share's own OpenAPI fragment declares HTTP Basic ("scheme": "basic"),
but the same Authorization: Bearer <token> used everywhere else works there
too (a structural 400 on a garbage body, not a 401).

Two distinct "Update Secret Note" routes, not a duplicate: the catalog lists
two operations with the same display name and different ids
(DOPPLER_UPDATE_SECRET_NOTE -> POST /v3/projects/project/note, and
DOPPLER_SECRETS_UPDATE_NOTE -> POST /v3/configs/config/secrets/note). The
CLI's own source labels the second one deprecated in favour of the first, but
confirmed live this session to still be functionally live (a structural 400,
not a route-absent 404) - both are genuinely real, distinct routes, so both
are implemented.

Rate limits: real, documented per-minute limits by bucket (reads,
secret-reads, writes are separate buckets) rather than a single number -
secret-reads at 120/min is the tightest and the one the live test suite paces
against.

Privacy-sensitive fields confirmed live, all handled explicitly rather than
left to a schema to catch: a Doppler Share link's password (the link's actual
decryption key, returned once in plaintext), a service token's key (the
full usable credential, returned once), and real user names/emails embedded
per-entry in workplace-user and activity-log list responses.

Plan gating: change requests and groups both answer a confirmed live 403 on
a Developer-plan account ("requires a Team or Enterprise plan" / "You do not
have access to groups"), matching the catalog's own stated gate for the
change-requests operation.

Agam00:
## Description

Adds a Doppler integration covering all 62 operations listed in the OSS
catalog: workplace settings and its users/roles/permissions, activity logs,
projects, project roles/permissions/members, environments, branch configs
(including clone/lock/unlock), config change logs (including rollback),
secrets (read/write/download/names/notes across two distinct note routes),
dynamic-secret lease revocation, service tokens, third-party integrations,
pending invites, group membership removal, webhooks, change requests, and
Doppler Share link creation (plain and end-to-end-encrypted).

Doppler publishes no single downloadable OpenAPI spec. This plugin was built
from `DopplerHQ/cli`'s own Go source (37 of 62 operations) plus the
per-operation OpenAPI 3.1 fragments embedded in `docs.doppler.com`'s
individual reference pages, discovered via the `llms.txt` index linked from
the site's `robots.txt` - the remaining 25 operations, cross-checked against
the CLI and live calls.

Fixes #796 

Docs: https://docs.doppler.com/reference/api
Catalog: https://corsair.dev/oss/doppler

## Operations

62 operations across 20 resource families:

| Family | Ops | Notes |
| --- | --- | --- |
| Workplace | 2 | get, update |
| Workplace users | 2 | list, get |
| Workplace roles | 3 | list, get, list permissions |
| Activity logs | 2 | list, retrieve |
| Projects | 5 | list, create, get, update, delete |
| Project roles | 3 | list, get, list permissions |
| Project members | 3 | list, get, delete |
| Environments | 5 | list, create, get, delete, rename |
| Configs | 8 | list, create, get, update, delete, clone, lock, unlock |
| Config logs | 3 | list, get, rollback |
| Secrets | 8 | list, get, delete, update, download, list names, update note x2 (see below) |
| Dynamic secrets | 1 | revoke lease |
| Service tokens | 3 | list, create, delete |
| Integrations | 1 | list |
| Invites | 1 | list |
| Groups | 1 | remove member |
| Webhooks | 7 | add, list, get, update, delete, enable, disable |
| Change requests | 1 | list (Team/Enterprise plans only) |
| Share | 2 | create plain-text link, create end-to-end-encrypted link |
| Auth | 1 | get authenticated user info |

Two catalog ids share the display name "Update Secret Note" but are two
genuinely distinct routes, not a duplicate:

- `DOPPLER_UPDATE_SECRET_NOTE` -> `POST /v3/projects/project/note`, the
  current, publicly documented route.
- `DOPPLER_SECRETS_UPDATE_NOTE` -> `POST /v3/configs/config/secrets/note`,
  not in the current public docs at all - only in the CLI, whose own source
  labels it deprecated in favour of the route above. Confirmed still live
  this session (a structural 400, not a route-absent 404), so it is
  implemented as the real, distinct operation the catalog lists it as, not
  collapsed into the first.

## Auth and transports

Single credential across both transports this plugin uses: a Bearer token,
sent as `Authorization: Bearer <token>` on every request. Confirmed live:
Doppler Share's own OpenAPI fragment for `/v1/share/*` declares HTTP Basic
(`"scheme": "basic"`), but the same Bearer token used on the documented `/v3`
API works there too - a structural 400 on a garbage body, not a 401. So this
is a second base URL (`api.doppler.com/v1/share` vs `api.doppler.com/v3`),
not a second auth code path.

Rate limiting is real, documented per-minute limits by bucket - reads,
secret-reads, and writes are separate buckets, confirmed from the docs
rather than assumed. `secret reads` at 120/min is the tightest and is what
the live test suite paces against. `retry-after` is honoured when Doppler
sends one.

## Persistence

Five entities mirrored: projects, environments, configs, webhooks, and the
workplace singleton. Only the primary key is required on every entity;
everything else is `.nullable().optional()`, and every object is `.loose()`.

Two identifier quirks worth naming, both handled explicitly rather than left
to `.loose()` to paper over:

- A **config**'s `name` is only unique *within* a project - `dev` exists in
  every project - so the local mirror keys every config cache/evict call on
  a composite `project:name` id rather than `name` alone, which would
  collide across projects.
- A **project**'s addressing param (`project=<slug>`, used by every other
  family too) is its `slug`, not its opaque `id` - the mirror keys on `slug`
  for the same reason: it is the one value every operation, including
  delete, actually has.

- **Secrets, in any shape.** Every secret-reading route returns the live
  value (`raw`/`computed`), not a masked fragment - there is no
  partial-exposure version of this data to cache.
- **Service tokens and Doppler Share links.** Both return a live, one-time
  credential (`key`, `password`) in their creation response. No entity
  exists for either, so there is no schema a future edit could accidentally
  widen to capture the credential field.
- **Activity logs, config logs, dynamic-secret leases.** Transactional,
  appended continuously, meaningful only against a time range.
- **Workplace users, project members, groups, invites, roles and
  permissions.** Identity/access data, not configuration - several of these
  families (groups, change requests) are plan-gated on the development
  account this plugin was built against, so their shape is declared from the
  spec rather than a live capture.

## Privacy

- **Doppler Share's `password` field is the link's actual decryption key, in
  plaintext, returned once.** Never logged, never mirrored - `auditPayload`
  on both `share.createPlain` and `share.createEncrypted` receives no field
  derived from the response.
- **A service token's `key` is the full, usable credential, returned once at
  creation.** Same treatment.
- **Workplace-user list/get and activity-log list/retrieve responses embed
  the real name and email of an account or the acting user, per entry.**
  Never logged - confirmed by a dedicated test that plants a real-shaped
  name/email in a mocked response and asserts neither reaches the event log.
- **A config log's `diff` embeds the actual before/after secret values that
  changed.** `configLogs.get` never passes it to `auditPayload`.
- **`secrets.update` logs the names of secrets it wrote, never the values.**
  A dedicated test plants a secret value in both the request and the mocked
  response and asserts it never appears in any logged payload, while the
  written *names* do.
- **A webhook's `authentication` is stripped before it reaches the local
  mirror**, even though confirmed live that Doppler only ever echoes back
  `{type}`, never the token/password itself - the strip costs nothing and
  keeps the mirror safe if that ever changes. The full record, including
  `authentication`, still reaches the caller.
- No real credential, workplace id, project slug, email, or personal
  identifier from the development account appears anywhere in this diff.
  Every fixture is fictional, verified with a self-tested scanner (real
  planted leaks all caught, a clean fixture stayed clean) run against the
  final diff.

## Tests

114 unit tests across 4 suites, plus a 9-test live suite excluded from CI.

- `endpoints.test.ts` (79 tests) - every one of the 62 operations: the
  route, method, and base URL it calls, with the request body asserted for
  every one of the 21 operations that sends one. A coverage sweep asserts
  the operations exercised are precisely the 62 registered. Plus 8 mirroring
  tests and 7 dedicated privacy tests (above).
- `schema.test.ts` (12 tests) - every live-captured key is declared against
  every `.loose()` entity, primary-key-only parsing, and a pinned assertion
  that no `secrets` or `serviceTokens` entity exists.
- `client.test.ts` (10 tests) - base URL and Bearer auth per transport, query
  serialisation, that a body is sent on DELETE (several Doppler routes
  address their target in a DELETE body, not the query string) as well as
  POST, and `retry-after` seconds-to-milliseconds conversion.
- `error-handlers.test.ts` (13 tests) - every handler classified by status
  first, message-text only as the fallback for a bare `Error`, including a
  regression pair proving a 500 whose body happens to mention "forbidden" or
  "not found" is not misclassified.
- `integration.test.ts` (9 tests) - live, self-skipping without credentials,
  paced at one request per 700ms against Doppler's documented 120/min
  secret-read limit. Read-only except a create-then-delete config probe and
  a Share link created with the default 1-view/1-day expiry so it is
  effectively spent by the time the test returns.

The live suite caught a real defect no mock could: `client.ts` stripped the
request body on every DELETE, silently breaking `projects.delete`,
`serviceTokens.delete`, and `dynamicSecrets.revokeLease` - all three address
their target in a DELETE body per Doppler's own spec, not the query string.
A transport-level test now pins that DELETE carries a body when one is
supplied. Also caught this way: `workplaceUsers.get`'s response is wrapped in
`{workplace_user: {...}}`, not flat as first assumed; `webhooks`' four
single-record routes are wrapped in `{webhook: {...}}` despite the docs'
example being empty `{}` for all of them - confirmed by creating, patching,
and deleting a real throwaway webhook live; and `changeRequests.list`'s
response is a bare JSON array, not the `{change_requests: [...], page}`
envelope every other paginated route uses - confirmed from the spec
fragment's own response schema, not assumed by analogy.

## Checklist

## Screenshots / Demos

image

## Additional Notes

**Footprint.** `packages/doppler/` (39 files, 36 TypeScript) plus a
three-line addition to `packages/corsair/core/constants.ts`, no deletions.

**No webhooks as triggers.** The catalog lists 0 triggers. Doppler does have
its own outbound webhook resource, and all 7 of its management operations
(add/list/get/update/delete/enable/disable) are among the 62 - but they are
exposed as ordinary managed-resource operations the agent calls, not as an
inbound event subscription Corsair reacts to.

**`secrets.update`'s spec-alternative `change_requests` body is out of
scope.** The spec accepts a `change_requests` array as a mutually-exclusive
alternative to the plain `{name: value}` map - conditional writes keyed on
an expected prior value, with per-field promote/delete/converge flags and
`valueType` validation. Not implemented: the catalog's own description for
this operation only asks for the plain map, and conditional writes are a
materially larger surface no catalog operation calls for.

**Operations confirmed live vs. mapped from source and spec.** The read
surface (workplace, projects, environments, configs, secret names, auth) and
a config create/delete/list round-trip were confirmed live this session,
along with a throwaway webhook create/patch/delete and a self-expiring Share
link. Group membership removal and change-request listing are implemented
and covered by mocked tests only - both confirmed to answer a real, live 403
plan-gate on this Developer-plan account rather than fired for real, since
doing so would spend a request to learn nothing new on every future run.

`corsair/http` or any other file outside `packages/doppler/`.

## Summary by CodeRabbit

* **New Features**
  * Added Doppler as a supported provider.
  * Added access to workplaces, projects, environments, configurations, secrets, webhooks, service tokens, integrations, and sharing links.
  * Added authentication, activity logs, change requests, groups, invitations, roles, and dynamic-secret lease management.
  * Added secure API communication, sanitized audit logging, and local persistence for supported resources.
* **Bug Fixes**
  * Added automatic rate-limit retries and clearer authentication, permission, and not-found errors.


Agam00:
@greptile review

Agam00:
@greptile review

Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
Tested locally LGTM

devjain32 (approved):

PR #800, opened 2026-08-16
feat(bigmailer): add BigMailer integration (57 ops)

### LINKED ISSUE #799: Integration request: BigMailer

BigMailer

### API documentation link

https://docs.bigmailer.io/reference

BigMailer is a multi-brand email marketing platform: an account holds brands,
and almost everything else - contacts, lists, segments, custom fields, merge-
tag properties, templates, senders, delivery connections, suppression lists
and both kinds of campaign - lives scoped underneath a brand. That shape is
the whole point of the product (agencies and multi-property companies run
dozens of brands off one account with separate sending identities), and it is
also what makes it tedious by hand: answering "is this brand ready to send"
means checking a verified sender, a delivery connection, a list, a template
and a campaign's ready flag across five different screens, per brand.

I would like an agent to be able to drive that directly, covering the 57
operations the catalog lists:

- Brands: list, create, get, update (4)
- Brand properties: list, create, get, update, delete (5)
- Custom fields: list, create, get, update, delete (5)
- Contact lists: list, create, get, update, delete (5)
- Connections: list (1)
- Message types: list (1)
- Senders: list (1)
- Contacts: list, create, get, update, delete, upsert, create batch,
  get batch status (8)
- Segments: list, create, get, update, delete (5)
- Suppression lists: list, create (CSV upload), get (3)
- Templates: list, create, get, update, delete (5)
- Bulk campaigns: list, create, get, update (4)
- Transactional campaigns: list, create, get, update (4)
- Users: list, create, get, update, delete (5)
- Auth: get authenticated user info (1)

The concrete case that motivated it: an agent can answer "which of our brands
can actually send right now" by reading senders, connections and lists per
brand, and then act on the answer - creating the missing custom field a
template's merge tag needs, upserting a batch of contacts onto the right
list, building a segment from campaign activity, and creating a bulk campaign
left deliberately un-ready so a human still presses send - without anyone
clicking through five separate BigMailer screens per brand to piece it
together.

_No response_

### Additional context

Source of truth: BigMailer publishes no single downloadable OpenAPI spec.
Their docs are ReadMe.io-hosted, and every reference page has a machine-
readable twin at docs.bigmailer.io/reference/<slug>.md, indexed by the site's
own llms.txt. Each of those pages carries the exact method, path, parameter
table and request/response schema for one operation - all 57 were read from
there rather than inferred from sibling endpoints.

Four resource families (contacts, segments, suppression lists, templates)
were then put through a second verification pass, which caught four real
mistakes an inference-by-analogy build would have shipped: contacts take
three independent *_op parameters (field_values_op, list_ids_op,
unsubscribe_ids_op), not one combined op; upsert contact is its own
POST /contacts/upsert path, not a variant of the by-id route; suppression
list upload is multipart/form-data with a binary file part, not a JSON body
with a base64 string; and segments' and templates' updates are POST, not the
PUT/PATCH used elsewhere in the same API.

Auth: a single account-level API key, sent as the X-API-Key header on every
request. One credential, one transport, one base URL
(https://api.bigmailer.io/v1) - no OAuth, no per-brand credential, no second
host. Matches the catalog's "1 auth".

Rate limits: documented, and unusually specific - 10 API calls per second and
a maximum of 4 concurrent calls per account, both stated on
docs.bigmailer.io/docs/getting-started-api, with HTTP 429 on breach. The docs
name no provider-specific retry-after-style header, so a client should honour
the standard retry-after if one arrives and otherwise use its own backoff
rather than inventing a header name.

Privacy-sensitive fields worth naming up front: contact email addresses are
personal data and are also the operation identifier on several routes (the
contact path parameter accepts an email in place of a UUID), so they need
handling as sensitive even where they look like a plain id; and a brand's
logo is a base64-encoded image that has no business in an audit log on size
grounds alone. The API itself is well-behaved here - connection and sender
objects never echo back the underlying AWS SES credential, confirmed against
each family's own reference page.

Agam00:
## Description

Adds a BigMailer integration covering all 57 operations listed in the OSS
catalog: brands, brand properties, custom fields, contact lists, delivery
connections, message types, senders, contacts (including upsert and the
asynchronous 1,000-contact batch upload), segments, campaign suppression
lists, templates, bulk campaigns, transactional campaigns, account users, and
the `/me` authentication check.

BigMailer publishes no single downloadable OpenAPI spec. Their docs are
ReadMe.io-hosted, and every reference page has a machine-readable twin at
`docs.bigmailer.io/reference/<slug>.md`, indexed by the site's own
`llms.txt` - each carrying the exact method, path, parameter table and
request/response schema for one operation. All 57 were read from there rather
than inferred from sibling endpoints, and four families were re-verified a
second time (see "Corrections caught by the second pass" below).

Fixes #799

Docs: https://docs.bigmailer.io/reference
Catalog: https://corsair.dev/oss/bigmailer

## Operations

57 operations across 15 resource families:

| Family | Ops | Notes |
| --- | --- | --- |
| Brands | 4 | list, create, get, update (the catalog lists no brand delete) |
| Brand properties | 5 | list, create, get, update, delete |
| Fields | 5 | list, create, get, update, delete |
| Lists | 5 | list, create, get, update (rename), delete |
| Connections | 1 | list |
| Message types | 1 | list |
| Senders | 1 | list |
| Contacts | 8 | list, create, get, update, delete, upsert, create batch, get batch status |
| Segments | 5 | list, create, get, update, delete |
| Suppression lists | 3 | list, create (multipart CSV upload), get |
| Templates | 5 | list, create, get, update, delete |
| Bulk campaigns | 4 | list, create, get, update (the catalog lists no campaign delete) |
| Transactional campaigns | 4 | list, create, get, update |
| Users | 5 | list, create, get, update, delete |
| Auth | 1 | get authenticated user info (`GET /me`) |

Everything except brands, users and `auth.me` is brand-scoped through a
`brand_id` path parameter - taken from each operation's own reference page
parameter table, not assumed from the family above it.

A dedicated coverage test asserts the operations the test suite exercises are
precisely the 57 the plugin registers, so a route added without a test, or a
test for a route that no longer exists, fails CI rather than passing quietly.

### Corrections caught by the second pass

Contacts, segments, suppression lists and templates were re-verified against
their own reference pages after the rest of the plugin was built. Four things
an inference-by-analogy build would have shipped wrong:

- **Contacts take three independent `*_op` parameters**
  (`field_values_op`, `list_ids_op`, `unsubscribe_ids_op`), each choosing
  add/replace/remove for its own collection - not one combined `op` covering
  the whole update.
- **Upsert is its own path**, `POST /brands/{brand_id}/contacts/upsert`, not a
  variant of the by-id route the other contact operations use.
- **Suppression list upload is `multipart/form-data`** with a binary `file`
  part, not a JSON body carrying a base64 string.
- **Segments' and templates' update routes are `POST`**, not the `PUT`/`PATCH`
  used elsewhere in the same API.

Each correction is documented at its own declaration in
`endpoints/types.ts` with what was wrong and how it was confirmed.

## Auth and transport

One credential, one transport, one base URL. An account-level API key sent as
`X-API-Key` on every request against `https://api.bigmailer.io/v1`. No OAuth,
no per-brand credential, no second host - matching the catalog's "1 auth".

Rate limiting is documented and specific rather than inferred: 10 API calls
per second and a maximum of 4 concurrent calls per account, both stated on
`docs.bigmailer.io/docs/getting-started-api`, with HTTP 429 on breach. The
docs name no provider-specific rate-limit headers, so `client.ts` configures
only the HTTP-standard `retry-after` - honoured when BigMailer sends one, with
the shared transport's exponential backoff (3 retries, 1s initial, 2x) as the
fallback rather than a guessed header name.

`error-handlers.ts` classifies purely by HTTP status (429/401/403/404 plus a
default), never by scanning a response body: unlike some providers, BigMailer
publishes no error-body schema anywhere in its reference docs, so matching on
message text would be matching against a shape that could be anything. The
message-text branch exists only as a fallback for a bare `Error` that carries
no status at all.

## Persistence

Thirteen entities mirrored: brands, brand properties, fields, lists,
connections, message types, senders, contacts, segments, suppression lists,
templates, bulk campaigns, transactional campaigns. Only the primary key is
required on every entity; everything else is `.nullable().optional()`, and
every object is `.loose()`. Fields come verbatim from each endpoint's own
reference page - the full documented response, not a hand-picked subset.

One identifier quirk drives the whole caching design, and is handled
explicitly rather than left to collide:

- **A brand-scoped id is only unique within its brand.** Two brands can each
  hold a list, a field, or a template with the same id-shaped key, so every
  brand-scoped cache and evict call keys on a composite `brand:id`, never the
  bare id. Brands themselves - the only top-level resource that is mirrored -
  key on their bare id. Tests pin both halves of that split, including that an
  evict uses the exact composite key the matching cache call wrote.

- **Account users.** Identity/access records describe *who* can act, not
  *what* is configured - a different kind of data than this reference-data
  mirror is for. A test pins that no user is ever cached.
- **Contact batches.** A transient processing status for an in-flight upload,
  not durable configuration. Also pinned by test.
- **Engagement aggregates as their own resource.** `engagement` on brands and
  lists, and the `num_*` counters on both campaign kinds, are continuously
  recomputed. They are still captured as fields on their entities - never
  dropped - just understood as a snapshot, not broken out into a cached
  resource of their own.

Mirroring is best-effort throughout: a failed local write logs a warning and
never fails the caller's API call. The one exception is eviction marked
`required: true`, where leaving the row behind would breach something the
plugin promises rather than merely leave a cache stale.

## Privacy

- **A contact's email address is personal data *and* an operation
  identifier.** BigMailer's contact routes document their path parameter as
  accepting an email address in place of a UUID, so a deny-list keyed only on
  the literal name `email` would miss it arriving as `contactId`. Both are
  covered: `contacts.ts` never passes `contactId` as an identifier key, and
  `email` is independently deny-listed by name in `logging.ts`. A test plants
  a real-shaped address and asserts it reaches neither audit payload.
- **A brand's `logo` is a base64-encoded image**, deny-listed on size and
  hygiene grounds - an audit log has no business carrying an image. Tested
  with a logo explicitly supplied on the call.
- **A brand property's value is never logged**, only its id - the property is
  the merge-tag content a brand injects into every email, and there is no
  reason an event log needs a copy of it.
- `key`, `secret`, `token`, `password`, `api_key` and `apikey` are deny-listed
  ahead of need. No documented BigMailer response returns any of them today -
  connection and sender objects never echo the underlying AWS SES credential,
  confirmed against each family's own reference page - so this is a second,
  independent guarantee that stays correct if that ever changes.
- No real credential, account id, brand id, email or personal identifier
  appears anywhere in this diff. Every fixture is fictional.

## Tests

89 unit tests across 4 suites, all passing, plus a live suite excluded from CI.

- `endpoints.test.ts` (75 tests) - all 57 operations, each asserting the exact
  method and path it calls, plus a coverage sweep pinning that the exercised
  set is precisely the 57 registered and that every delete is marked
  `destructive`. Then 8 mirroring tests (bare-id vs composite `brand:id`
  keying, list-wide caching, evict-key correctness, and the two deliberate
  non-caches), 3 privacy tests (above), and 5 request-body tests - including
  that field `type` cannot be changed through an update, that `ready` is
  omitted entirely on campaign create unless the caller explicitly asks for
  it, and that a suppression list is sent as real multipart form data decoded
  from the base64 input.
- `client.test.ts` (6 tests) - base URL, `X-API-Key` header, that an empty or
  whitespace key throws before any request is issued, query serialisation, and
  the rate-limit config's shape.
- `error-handlers.test.ts` (6 tests) - each handler classified by status
  first, with message text used only as the fallback for a bare `Error`.
- `schema.test.ts` (2 tests) - every entity parses from its primary key alone
  and preserves unknown keys through `.loose()`.
- `integration.test.ts` (1 test) - live, self-skipping unless
  `BIGMAILER_API_KEY` is set, currently a read-only `auth.me` +
  `brands.list` probe.

## Checklist

## Screenshots / Demos
image

## Additional Notes

**Live verification status - please read before reviewing.** No BigMailer API
key was available while this plugin was built, so every route here is mapped
from BigMailer's own per-operation reference pages and verified by mocked
tests, not confirmed against the live API. The `docs.bigmailer.io/reference/
<slug>.md` pages were fetched live and each carries the operation's real
method, path and schema, so this is documentation-grounded rather than
guessed - but it is not the same standard as a live round-trip, and I am not
claiming it is. The live suite in `integration.test.ts` is deliberately kept
as a runnable stub for exactly this reason: it skips itself without
`BIGMAILER_API_KEY`, and extending it into a full create/update/delete cycle
is the first thing that should happen once an account is available. R4's demo
recording will be added before this leaves draft.

**Two specifics the docs could not settle**, each documented at its call site
rather than silently chosen:

- **`transactionalCampaigns.update`'s HTTP method.** Repeated fetches of the
  operation's own reference page never surfaced the method unambiguously.
  `PATCH` is used, matching `bulkCampaigns.update` - the closest sibling in
  the same family, on the same resource shape. Flagged for live confirmation.
- **`users.create`'s path.** `POST /users` is used, consistent with the other
  four user operations, which are all unscoped `/users` routes. A docs
  summarizer suggested `/accounts/{account_id}/users`, which no reference page
  corroborated and which would be the only account-scoped path in the whole
  API surface. Flagged for live confirmation.

Both are cheap to correct if the live check disagrees - a single method or
path string each, with their tests pinning the change.

**Footprint.** `packages/bigmailer/` (34 files, 31 TypeScript) plus a
three-line registration in `packages/corsair/core/constants.ts` and the
generated `pnpm-lock.yaml` entry. No deletions, nothing else touched - R1
scope exactly.

**No webhooks, no triggers.** The catalog lists 0 triggers, and BigMailer's
public REST API publishes no webhook or event-subscription resource of any
kind - there is not even an outbound-webhook management surface to expose as
ordinary operations. The plugin has no `webhooks/` directory and empty
`webhooks`/`webhookSchemas` maps.

**Deliberately out of scope:** nothing in the catalog's 57 operations is
partially implemented or stubbed. Where BigMailer's docs publish no formal
schema for a nested payload - a contact's per-field value object, a segment's
`conditions` array, whose shape varies by condition type and is shown only as
two worked examples - the plugin keeps it `.loose()` with optional members
rather than over-modelling a shape from partial documentation.

`corsair/http` or any other file outside `packages/bigmailer/`.

## Summary by CodeRabbit

- **New Features**
  - Added BigMailer integration with API-key authentication and 57 supported operations.
  - Manage brands, contacts, lists, fields, segments, templates, senders, suppression lists, users, and campaigns.
  - Added batch contact processing, campaign scheduling, pagination, filtering, and suppression-list uploads.
  - Added local caching, typed validation, and privacy-conscious operation logging.
- **Bug Fixes**
  - Added automatic retries for rate-limited requests, including server-provided retry delays.
  - Improved handling of authentication, permission, missing-resource, and other API errors.


Agam00:
@greptile review

devjain32 (approved):

PR #802, opened 2026-08-16
fix(telegram): use constant-time comparison for webhook secret token

### LINKED ISSUE #707: fix(telegram): use constant-time comparison for secret token
## What's wrong

`packages/telegram/webhooks/types.ts` still checks the bot secret token with `providedToken === secretToken`.

## Change

Swap that compare for `crypto.timingSafeEqual` after a length check. Import `crypto` from `node:crypto` if needed. Keep the existing invalid error on failure.

Leave the empty-secret guard alone. That path already fails closed.

## Check

Matching token still valid. Mismatch and length mismatch still invalid. There is already `packages/telegram/webhooks/webhooks.test.ts`; add cases there rather than a new file.

wrvnnull:
## Description

Fixes #707.

`verifyTelegramWebhookSignature` compared the webhook secret with `===`, which can leak the token via timing. This switches to a UTF-8 byte-length guard plus `crypto.timingSafeEqual`.

Tokens are encoded to buffers first so a non-ASCII value that matches the JS string length but not the byte length is rejected without `timingSafeEqual` throwing.

The empty-secret guard is unchanged (already fails closed). Regression tests cover ASCII length mismatch and Unicode byte-length mismatch.

## Checklist

Webhook secret comparison is unit-tested (no live Telegram call). Passing proof:

- Change: https://github.com/corsairdev/corsair/blob/de3d590a03033bbe4861593e06ecce3028e4cc06/packages/telegram/webhooks/types.ts
- Tests: https://github.com/corsairdev/corsair/blob/de3d590a03033bbe4861593e06ecce3028e4cc06/packages/telegram/webhooks/webhooks.test.ts
- CI Checks: https://github.com/corsairdev/corsair/actions/runs/31949420627

## Additional Notes

This is the spec'd bug fix from #707 (not a new integration).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook secret validation to safely reject tokens with mismatched lengths without errors.
  * Added coverage for invalid token formats and character encodings.

* **Refactor**
  * Removed duplicate provider entries while preserving existing provider names and labels.


yuvrxj-afk:
The byte-length guard is already in place: both tokens are encoded to UTF-8 buffers and their byte lengths compared before `timingSafeEqual`, so a non-ASCII token with a matching string length but different byte length is rejected as invalid rather than throwing. This looks addressed already.

@greptileai review

devjain32 (approved):

PR #803, opened 2026-08-16
feat(frpc): ship frpc as per-platform npm packages

yuvrxj-afk:
## What

Deliver `frpc` as per-platform npm packages (`@corsair-dev/frpc-<platform>-<arch>`) — the esbuild/Turbopack model — so `corsair` can resolve the binary from the **npm registry** instead of the postinstall GitHub download.

- 6 packages: `darwin-{x64,arm64}`, `linux-{x64,arm64}`, `win32-{x64,arm64}`, each with `os`/`cpu` fields so the package manager installs **only** the matching one.
- `corsair` lists all 6 as `optionalDependencies` (`workspace:*`, version-locked).
- `resolveFrpcBinary()` now prefers the installed platform package, falling back to the existing postinstall cache.
- `scripts/prepare-frpc-packages.mjs` materializes each binary at publish (SHA256-verified against the pinned checksums; frp's Apache-2.0 LICENSE + a NOTICE bundled). Binaries are git-ignored — present only in the published tarball.

## Why

pnpm blocks dependency postinstall scripts by default (the `approve-builds` friction). A package with no install script sidesteps that entirely, and the binary comes from the registry (cached, offline-friendly) rather than a GitHub hop.

**Postinstall is intentionally kept** as a fallback this round; a follow-up removes it once the packages are proven in the wild.

## Verified locally
- `prepare-frpc-packages.mjs` downloads + SHA256-verifies + extracts all 6 (binary + frp LICENSE).
- `npm pack` includes the binary (the `files` allowlist wins over the repo `.gitignore`).
- **os/cpu gating**: installing all 6 tarballs into a clean project added **only** `frpc-darwin-arm64` on this Mac (1 of 6).
- `resolveFrpcBinary()` resolves the package binary; SDK typechecks + builds (createRequire survives ESM bundling).

## Notes
- Based on `feat/frp-dev-tunnel` (#770) since it needs `frpc-binary.ts`. After #770 merges, rebase onto main + retarget.
- Frp is Apache-2.0; redistributing the binary requires shipping its LICENSE + attribution, which each package does.

## Summary by CodeRabbit

* **New Features**
  * Added automatic support for prebuilt tunnel components across macOS, Linux, and Windows on ARM64 and x64 systems.
  * Improved platform detection and fallback handling when starting development tunnels.
  * Added clearer tunnel output with the live tunnel link and delivery guidance.

* **Bug Fixes**
  * Improved download verification and fallback reliability for tunnel components.
  * Simplified tunnel startup messaging and Ctrl+C instructions.

* **Documentation**
  * Added platform-specific information and licensing notices for bundled tunnel components.


yuvrxj-afk:
@greptileai

devjain32 (approved):

yuvrxj-afk:
@greptileai

PR #807, opened 2026-08-16
feat(bigml): implement BigML plugin with API client

### LINKED ISSUE #806: Integration request: BigML

BigML

### API documentation link

https://bigml.com/api

BigML is a machine-learning platform: an account holds projects, and every
computed resource - sources, datasets, models, predictions, clusters,
anomaly detectors, and a dozen other resource types - is organized under
them. The catalog for this integration covers 45 operations scoped to
project/source management plus read access across the platform's resource
types:

- Projects: create, get, delete, list (4)
- Sources: get, update, list (3)
- External connectors: create, get (2)
- Configurations: get, list (2)
- Read/list access across 19 computed resource types: anomalies, anomaly
  scores, batch anomaly scores, associations, association sets, batch
  centroids, batch predictions, batch projections, batch topic
  distributions, centroids, clusters, composites, deepnets, ensembles,
  evaluations, executions, forecasts, fusions, libraries, linear
  regressions, logistic regressions, models, optiml, PCAs, predictions,
  projections, samples, scripts, statistical tests, time series, topic
  distributions, topic models (34)

I would like an agent to be able to inventory and manage the project/source
layer of a BigML account directly - creating a project, connecting an
external data source, and then listing and inspecting the models, datasets,
and prediction resources that already exist under it - without switching to
the BigML dashboard.

I have claimed this integration on corsair.dev/oss and am building the
plugin.

_No response_

### Additional context

Source of truth: BigML's own docs site (bigml.com/api) is a JavaScript-
rendered SPA with no static/downloadable spec. Ground truth for this build
comes from BigML's official Python SDK (github.com/bigmlcom/python, 293
stars), which encodes every resource path, HTTP verb, and the auth mechanism
directly in code - confirmed base domain bigml.io, API version segment
andromeda, and the following per-resource pattern read from
bigmlconnection.py:

- List: GET {resource}, response envelope {meta, objects}
- Create: POST {resource}, body JSON
- Get: GET {resource}/{id}
- Update: PUT {resource}/{id} (confirmed PUT, not the POST-everywhere
  convention some other providers use - read from the SDK's actual
  requests.put(...) call, not assumed)
- Delete: DELETE {resource}/{id}, 204 No Content on success

Auth: a username + api_key pair, both required, sent as query-string
parameters on every request (not a header) - confirmed from the SDK's
_add_credentials method. No OAuth, no per-project credential.

Rate limiting: plan-tiered, returns 429 with a JSON error body; BigML
documents no custom rate-limit header, so the client honours the standard
retry-after if present and otherwise falls back to its own backoff.

One thing flagged for live confirmation rather than assumed: LIST_COMPOSITES
is very likely not a separate resource path - the SDK creates a composite
source by POSTing to the plain /source endpoint with a sources: [...] array
field, so composites are probably a filtered view of /source rather than
their own route. This gets confirmed against a live account before shipping,
not guessed.

Async resources: many BigML resources (datasets, models, predictions, etc.)
are computed asynchronously - creation can return 202 Accepted while BigML
finishes processing in the background, and a resource's status.code field
tracks progress. This catalog's scope is deliberately limited to
project/source management plus read/list access, which sidesteps needing to
model that async lifecycle for a synchronous tool call - a narrower but more
honest scope than exposing create/train operations for resources whose
completion this integration would then have no way to wait for.

Agam00:
## Description

Adds a BigML integration covering all 45 operations listed in the OSS
catalog: project management (create/get/delete/list), source management
(get/update/list), external data connectors (create/get), saved
configurations (get/list), and read/list access across 34 of BigML's
computed-resource types (models, datasets, predictions, clusters, anomaly
detectors, and more - see "Operations" below for the scope note).

BigML's docs site (bigml.com/api) is a JavaScript-rendered SPA with no
static/downloadable spec. Ground truth for routing came from BigML's
official Python SDK (`bigmlcom/python`, 293 stars), which encodes every
resource path, HTTP verb, and the auth mechanism directly in code - and
every route, field shape, and edge case from the SDK was then re-confirmed
against a real live account before being trusted, including two corrections
the SDK's own docstrings would not have caught (see "Live corrections"
below).

Fixes #806 

Docs: https://bigml.com/api
Catalog: https://corsair.dev/oss/bigml

## Auth and transport

`username` + `api_key`, both required, sent as query-string parameters on
every request - confirmed from the SDK's `_add_credentials` method and
verified live. No OAuth, no per-project credential. Base URL
`https://bigml.io/andromeda`.

Per-resource REST convention, confirmed live against a real account, not
assumed from the SDK alone:

- List: `GET {resource}` -> `{meta, objects}`
- Create: `POST {resource}`
- Get: `GET {resource}/{id}`
- Update: `PUT {resource}/{id}` (confirmed live: a no-op source rename
  returned `202 Accepted` - this API's real convention is `PUT`, not the
  `POST`-everywhere convention some other providers in this repo use)
- Delete: `DELETE {resource}/{id}` -> `204 No Content`

Every resource is keyed by its own `resource` field, a compound
`{type}/{hex24}` string that is globally unique across the account -
confirmed live, so no composite cache key is needed anywhere in this
plugin (unlike several other plugins in this repo whose ids are only
unique within a parent scope).

Pagination is `limit`/`offset`, not cursor-based - confirmed real by effect:
an invalid `limit` value 400s rather than being silently ignored, and
different `offset` values return different rows.

Rate limiting: plan-tiered `429` with a JSON error body; no documented
custom header, so the client honours the standard `retry-after` if present
and otherwise falls back to its own backoff. `402` (plan/task-limit) is
treated the same as `403` - both are non-retryable write rejections, not a
scope problem.

## Live corrections

Two things the SDK alone would have gotten wrong, both caught by testing
against a real account rather than trusting the SDK's docstrings:

- **`LIST_COMPOSITES` is a real, separate `/composite` resource path.**
  The SDK's own `_create_composite` method posts to the plain `/source`
  endpoint with a `sources: [...]` array, which looked like strong evidence
  that composites were just a filtered view of `/source` - flagged for live
  confirmation rather than assumed. The live account proved that guess
  wrong: `/composite` is its own top-level listing, distinct from `/source`.
- **External connector `connection` fields have a confirmed vocabulary,
  and the connector type is a top-level field, not nested.** BigML's own
  validation error on an invalid `connection` key names the exact accepted
  set (`host`, `hosts`, `port`, `database`, `use_ssl`, `verify_certs`,
  `user`, `password`, `http_auth`, `sslmode`, `master`, `timeout`,
  `indice`), and a `source: 'postgresql'` field sent at the top level
  (not inside `connection`) is what BigML actually expects.
- **`sources.update` was missing two things its own catalog description
  names**: "parsing configuration" and "field properties". A first pass
  scoped the operation to `name`/`description`/`tags` only; re-reading the
  catalog text caught the gap, and both are now modelled -
  `sourceParser` (`{separator, locale, missingTokens}`) and `fields` (a map
  from BigML's own field id to `{name, label, description, optype}`), both
  shapes taken from a real source's live response, not guessed. Confirmed
  live that both 400 with `"Cannot update closed source"` once a source has
  finished processing (true for essentially every source a caller would
  already have an id for) - documented at the schema, not silently dropped.
- **Every list operation was missing `orderBy`/filtering.** The catalog
  explicitly promises "filtering, ordering, and pagination" on most list
  descriptions; a first pass only implemented pagination. Confirmed live
  that `order_by=size`/`order_by=-size` genuinely sort (and an invalid
  field name 400s rather than being ignored), and that arbitrary
  `field=value` query params genuinely filter (confirmed by effect: a
  nonexistent-name filter returns `total_count: 0`). Both are now wired
  into every one of the 38 list operations through one shared `listQuery`
  helper (`endpoints/shared.ts`) rather than per-endpoint, so none of them
  can drift out of sync with each other.

## A real credential-exposure finding, and how it's handled

Two things confirmed live against a real account that this plugin has to
actively defend against, not just note:

- **BigML echoes a connector's `password`/`user` back in plaintext** on
  every subsequent `GET` and in the `LIST` envelope - this is BigML's own
  API behaviour, not a bug here. `externalConnectors` is therefore the one
  resource this plugin never mirrors locally (`schema/index.ts`), and
  `connection` is deny-listed by name in `logging.ts` so no audit event can
  carry it either - both independently, and both proven with a
  mutation-tested guard (`endpoints.test.ts`: the deny-list fault was
  planted, watched to fail the intended test, then reverted).
- **BigML's own `meta.next`/`meta.previous` pagination links embed the
  live account's `username`/`api_key` directly in plain text.** Confirmed
  live on `GET /source?limit=1`. `client.ts`'s `redactPaginationCredentials`
  strips both query params from every response before it reaches a caller,
  applied centrally so no individual endpoint can forget it - also
  mutation-tested.

## Operations

45 operations. Scope is deliberately project/source management plus
read/list access across the platform's computed-resource types - no
create/train operations for datasets, models, or predictions. Those
resources are computed asynchronously in BigML (`create` can return `202`
while a background job runs, tracked via a `status.code` lifecycle
confirmed live even on `project` and `source`), and this catalog's scope
does not attempt to model that lifecycle for a synchronous tool call.

| Family | Ops | Notes |
| --- | --- | --- |
| Projects | 4 | create, get, delete, list |
| Sources | 3 | get, update, list (no create - not in catalog scope) |
| External connectors | 2 | create, get |
| Configurations | 2 | get, list |
| 34 generic computed-resource types | 34 | list only - anomalies, datasets, models, predictions, clusters, and 29 more, each its own family (`anomalies.list`, `datasets.list`, ...) since they are independent resource types on the account, not variants of one another |

## Persistence

Entities keyed on their bare `resource` field. `projects`, `sources`,
`configurations` get their own typed entity, captured from live responses
(`GET /project`, `GET /source`) - not transcribed from docs, since BigML's
docs are unreadable without a browser. The 34 generic list-only types share
one conservative entity (the common envelope confirmed live on every
resource type checked: `resource`, `name`, `category`, `created`, `status`,
`tags`, and more) rather than fabricated per-type fields for resources this
account had no live examples of - a live pass against a populated account
is the natural next step to split these into per-type schemas.

`externalConnectors` is the one resource never cached - see "credential
exposure" above.

## Tests

84 unit tests across 4 suites, all passing, 71 assertions.

- `endpoints.test.ts` (59 tests) - all 45 operations, each asserting the
  exact method and path it calls; a coverage sweep pinning that the
  exercised set is precisely the 45 registered; caching tests including
  that external connectors are never cached; two privacy tests for the
  connection-credential deny-list (one mutation-tested); request-body
  tests for every write operation, including the new `sourceParser`/
  `fields` update surface; and a live-effect-confirmed test that `orderBy`
  and `filter` reach the query string correctly.
- `client.test.ts` (6 tests) - base URL, auth query params, that an empty
  username or api key throws before any request is issued, body-on-write
  behaviour, error wrapping, and the pagination-credential redaction
  (mutation-tested: the redaction function was neutered, the intended test
  confirmed to fail, then reverted).
- `error-handlers.test.ts` (7 tests) - each handler classified by status
  first, message text used only as the fallback for a bare `Error`, and an
  explicit proof that a status-bearing error is never message-sniffed even
  when its message contains another status's trigger word.
- `schema.test.ts` (12 tests) - every entity parses from `resource` alone,
  rejects a keyless record, preserves unknown keys through `.loose()`, and
  the entity registry is pinned to exactly the 37 stores this plugin uses
  (34 generic + projects/sources/configurations), with `externalConnectors`
  explicitly asserted absent.

## Checklist

## Screenshots / Demos

image

## Additional Notes

**No webhooks, no triggers.** The catalog lists 0 triggers, and BigML's
REST API has no webhook or event-subscription resource of any kind.

**Footprint.** `packages/bigml/` plus a three-line registration in
`packages/corsair/core/constants.ts` and the generated `pnpm-lock.yaml`
entry. No deletions, nothing else touched - R1 scope exactly.

**Deliberately out of scope, named explicitly:** create/train operations
for the 34 read-only computed-resource types (async lifecycle, not a fit
for this catalog); per-type field schemas for those same 34 types beyond
the common envelope (no live examples existed in the account this was
built against - flagged for a follow-up live pass rather than guessed).

## Summary by CodeRabbit

* **New Features**
  * Added BigML integration with authenticated API access, rate-limit retries, and clear error handling.
  * Added support for projects, sources, external connectors, configurations, and numerous resource listings.
  * Added resource caching, pagination, filtering, audit logging, and protection for sensitive credentials.
  * Added BigML provider support and package configuration.

* **Tests**
  * Added comprehensive coverage for API requests, endpoint operations, schemas, authentication, retries, error classification, and integration workflows.


Agam00:
@greptile review

Dhirenderchoudhary:
@greptile check

Dhirenderchoudhary:
Tested locally lgtm

devjain32 (approved):

PR #825, opened 2026-08-17
fix(corsair): return successful retry result instead of rethrowing

Mayank-saraswal:
## Description

Fixes a bug in the endpoint retry loop (`packages/corsair/core/endpoints/bind.ts`) where a **successful retry still threw the original error**.

### Root cause

Inside the `call()` retry function, the recursive retry was dispatched with `await call(newAttempt, ...)` and its result was discarded. After the retry ran, control fell through to an unconditional `throw error` at the end of the `catch`.

Trace for `call(0)`:

1. `value()` throws (e.g. a Slack 429).
2. `attemptNumber(0) < maxRetries` is true, so a retry is scheduled.
3. `call(1)` runs and this time `value()` **succeeds**, returning the result to the `call(0)` frame.
4. `call(0)` discards that result and hits `throw error` (the original 429).

So a retry that eventually succeeded always surfaced the original error to the caller. The retry feature never actually recovered a request. Slack and Linear ship a default `RATE_LIMIT_ERROR` handler with `maxRetries: 5`, so this path is reachable in normal use.

### Fix

`packages/corsair/core/endpoints/bind.ts`

```diff
 await new Promise((resolve) => setTimeout(resolve, delayMs));
-await call(newAttempt, callCtx, callArgs);
-
-console.log(
-	`[corsair:${pluginId}:${operationPath}] Retry strategy:`,
-	retryStrategy,
-);
+return await call(newAttempt, callCtx, callArgs);
```

The result of the retry is now returned, so a successful retry returns instead of falling through to `throw error`. The removed `console.log` would have become unreachable code after this change. The `throw error` at the end of the `catch` now only runs when no retry could be scheduled (i.e. retries are exhausted), which is the correct behavior.

### Test

Adds `packages/corsair/tests/retry-recover.test.ts`, which drives a real bound Slack endpoint through the retry loop:

- First attempt throws an `ApiError` 429.
- The retry succeeds and the caller now receives the successful result (asserts `makeSlackRequest` was called twice and the returned value is intact).
- A second case asserts that when every retry fails, the call still rejects with an error.

This test was verified both ways:

- With the buggy code it **fails** (the successful retry still rethrows the original error).
- With the fix it **passes** (2/2).

## Checklist

- [ ] I have added or updated necessary documentation

Terminal: the new test suite `tests/retry-recover.test.ts` passes 2/2 after the fix, and fails 1/2 before the fix.

## Additional Notes

- `pnpm lint` / biome: this change is a single TS logic fix. Running biome on this Windows checkout flags pre-existing CRLF line endings across the repo (including untouched files like `core/errors/handler.ts`); the new test file passes biome cleanly. This is an environment artifact, not introduced by this change.
- `pnpm typecheck` / `pnpm build` were not run across the whole monorepo (143 packages, heavy). The change compiled and ran under ts-jest in the test suite.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retry handling so successful retries return the expected response.
  - Ensured errors are correctly reported when all retry attempts fail.
  - Improved recovery from rate-limit errors during Slack requests.

- **Tests**
  - Added coverage for successful recovery after rate-limit errors.
  - Added coverage for failures after exhausting all retry attempts.
  - Verified that retried responses are returned correctly.

### DIFF
--- a/packages/corsair/core/endpoints/bind.ts
+++ b/packages/corsair/core/endpoints/bind.ts
@@ -242,12 +242,7 @@ export function bindEndpointsRecursively({
 								}
 
 								await new Promise((resolve) => setTimeout(resolve, delayMs));
-								await call(newAttempt, callCtx, callArgs);
-
-								console.log(
-									`[corsair:${pluginId}:${operationPath}] Retry strategy:`,
-									retryStrategy,
-								);
+								return await call(newAttempt, callCtx, callArgs);
 							}
 						}
 						throw error;
--- a/packages/corsair/tests/retry-recover.test.ts
+++ b/packages/corsair/tests/retry-recover.test.ts
@@ -1,0 +1,110 @@
+import { slack } from '@corsair-dev/slack';
+import { makeSlackRequest } from '@corsair-dev/slack/client';
+import { ApiError } from '../async-core/ApiError';
+import type { ApiRequestOptions } from '../async-core/ApiRequestOptions';
+import type { ApiResult } from '../async-core/ApiResult';
+import { createCorsair } from '../core';
+import { createTestDatabase } from './setup-db';
+
+jest.mock('@corsair-dev/slack/client', () => {
+	const original = jest.requireActual('@corsair-dev/slack/client');
+	return {
+		...original,
+		makeSlackRequest: jest.fn(),
+	};
+});
+
+const mockedMakeSlackRequest = makeSlackRequest as jest.MockedFunction<
+	typeof makeSlackRequest
+>;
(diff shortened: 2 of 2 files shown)


Mayank-saraswal:
@greptileai

Mayank-saraswal:
@devjain32  @yuvrxj-afk

ambikeesshh:
@greptileai

ambikeesshh (commented):
lgtm

devjain32 (approved):

PR #828, opened 2026-08-17
fix(teams): use constant-time comparison for clientState (#684)

### LINKED ISSUE #684: fix(teams): use constant-time comparison for clientState
## Problem
`verifyTeamsClientState` compares with `===`:

`packages/teams/webhooks/types.ts`:
```ts
);
```

Not constant-time. Different from #625 (empty `value[]`). This issue is only the string compare.

## Fix
Use `crypto.timingSafeEqual` with equal-length buffers.

Note: `packages/onedrive/webhooks/types.ts` and `packages/sharepoint/webhooks/types.ts` compare `clientState` the same non-constant-time way, so they are not a model to copy — they have the same bug.

## Touch
- `packages/teams/webhooks/types.ts` (add `import crypto from 'node:crypto'`)

## Done when
- [ ] `n.clientState === expectedClientState` is gone
- [ ] length mismatch and value mismatch both return invalid
- [ ] matching clientState still returns valid

## Verify
```bash
rg 'clientState === expectedClientState' packages/teams/webhooks/types.ts
rg timingSafeEqual packages/teams/webhooks/types.ts
```

chiliec:
## Description

Closes #684.

`verifyTeamsClientState` in `packages/teams/webhooks/types.ts` compared each notification's `clientState` with `===`, which is not constant-time and can leak timing information about the expected secret value.

This PR replaces the `===` comparison with `crypto.timingSafeEqual` after a length check, matching the pattern already used in `packages/vapi/webhooks/types.ts`:

```ts
const expected = Buffer.from(expectedClientState);
const allMatch = payload.value.every((n) => {
    const actual = Buffer.from(n.clientState ?? '');
    return actual.length === expected.length && timingSafeEqual(actual, expected);
});
```

The rest of the verifier is untouched — the existing `expectedClientState is required` guard stays, empty `value[]` behaviour is unchanged (that is #625, out of scope here), and all valid/invalid outcomes are preserved. Scope is a single plugin package (`packages/teams/`).

## Checklist

- [x] I have added or updated necessary documentation (no docs change needed — internal verifier behaviour is unchanged)

This is a security hardening change to an internal signature-verification helper, not a UI/CLI change, so there is no visual output to record. The behaviour is verified by unit tests in `packages/teams/webhooks/verify-client-state.test.ts` (7 assertions), which pass:

```
```

Because this is a **timing** fix rather than a behavioural one, the outcome-level tests pass against both the old `===` and the new `timingSafeEqual` implementation (verified by reverting locally) — they lock in that the constant-time rewrite preserves every valid/invalid result, while the security property itself comes from using `timingSafeEqual` with a length guard, mirroring the repo's own `vapi` verifier.

## Additional Notes

No new dependencies (`node:crypto` is built in). No breaking changes. Distinct from #625 (empty `value[]` handling).

### DIFF
--- a/packages/teams/webhooks/types.ts
+++ b/packages/teams/webhooks/types.ts
@@ -1,3 +1,4 @@
+import { timingSafeEqual } from 'node:crypto';
 import type { CorsairWebhookMatcher, RawWebhookRequest } from 'corsair/core';
 import { z } from 'zod';
 
@@ -195,9 +196,13 @@ export function verifyTeamsClientState(
 	if (!expectedClientState) {
 		return { valid: false, error: 'clientState is required' };
 	}
-	const allMatch = payload.value.every(
-		(n) => n.clientState === expectedClientState,
-	);
+	const expected = Buffer.from(expectedClientState);
+	const allMatch = payload.value.every((n) => {
+		const actual = Buffer.from(n.clientState ?? '');
+		return (
+			actual.length === expected.length && timingSafeEqual(actual, expected)
+		);
+	});
 	if (!allMatch) {
 		return { valid: false, error: 'clientState mismatch' };
 	}
--- a/packages/teams/webhooks/verify-client-state.test.ts
+++ b/packages/teams/webhooks/verify-client-state.test.ts
@@ -1,0 +1,68 @@
+import type { TeamsNotification, TeamsWebhookPayload } from './types';
+import { verifyTeamsClientState } from './types';
+
+// Regression for #684: verifyTeamsClientState compared clientState with `===`,
+// which is not constant-time. The comparison now uses timingSafeEqual after a
+// length check, while preserving the valid/invalid outcomes.
+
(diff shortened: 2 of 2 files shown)


Dhirenderchoudhary:
@greptile

chiliec:
The earlier "spans two packages" flag was from the pre-rebase diff — the branch is now rebased onto `main` and touches only `packages/teams/` (2 files), which the latest gate run confirms (R1 ✅). Checklist boxes are now all ticked (R3).

On R4 (demo video): this is a security-hardening change to an internal signature-verification helper (`===` → `crypto.timingSafeEqual`) with no UI/CLI output and no way to call the real Teams API from CI, so there's nothing to record. It's covered by 7 unit-test assertions in `verify-client-state.test.ts`. Happy for a maintainer to take the final review from here per the gate's note.

devjain32 (approved):

PR #888, opened 2026-08-21
fix(core): avoid quadratic backtracking in path parameter substitution

Dhirenderchoudhary:
## Description

`getUrl` substitutes path parameters with a lazy quantifier:

``.replace(/{(.*?)}/g, …)``

.*? backtracks from every { in the string, so the scan is quadratic in the
number of unmatched opening braces. CodeQL flags it as a polynomial regular
expression on uncontrolled data.

This replaces the lazy quantifier with an explicit negated class:

``.replace(/\{([^{}]*)\}/g, …)``

[^{}]* cannot cross a brace, so there is nothing to backtrack over and the
scan is linear. It matches exactly the same well-formed {placeholder} tokens.

## Impact

Every plugin request goes through getUrl, and options.url is built from
endpoint paths that interpolate caller-supplied values. A value carrying many
unmatched { characters stalls the event loop before the request is even sent.

## Tests

packages/corsair/tests/request-path-substitution.test.ts — 6 tests driving the
real request() with a stubbed fetch, asserting the resolved URL.

Five cover substitution behaviour so the change is provably semantics-preserving:
single placeholder, multiple placeholders, unmatched placeholder left untouched,
value encoding, and {api-version}.

The sixth is the regression guard: it builds the hostile input above and asserts
substitution completes in under 1s. With the lazy quantifier restored it fails
at 22,664 ms; the other five still pass, which is what confirms the fix changes
performance and not behaviour.

``pnpm --filter corsair test -- tests/request-path-substitution.test.ts``

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL path-parameter replacement for malformed inputs with many unmatched braces, preventing slowdowns.
  * Preserved correct substitution and encoding of valid path parameters, including API version placeholders.

* **Tests**
  * Added coverage for single and multiple parameters, empty and nested placeholders, unmatched braces, encoded values, API version handling, and large malformed inputs.

### DIFF
--- a/packages/corsair/async-core/request.ts
+++ b/packages/corsair/async-core/request.ts
@@ -91,7 +91,11 @@ const getUrl = (config: OpenAPIConfig, options: ApiRequestOptions): string => {
 
 	let path = options.url
 		.replace('{api-version}', config.VERSION)
-		.replace(/{(.*?)}/g, (substring: string, group: string) => {
+		// `[^{}]*` instead of a lazy `.*?`: the lazy form backtracks from every
+		// `{` in the string, which is quadratic on input like `"{a".repeat(n)`
+		// (~20s at n=200k). Excluding braces makes the scan linear and matches
+		// the same well-formed `{placeholder}` tokens.
+		.replace(/\{([^{}]*)\}/g, (substring: string, group: string) => {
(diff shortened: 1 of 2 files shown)


Dhirenderchoudhary:
@greptile

PR #890, opened 2026-08-21
fix(whatsapp): use constant-time comparison for hub.verify_token

### LINKED ISSUE #704: fix(whatsapp): use constant-time comparison for hub.verify_token
## What's wrong

`verifyWhatsappWebhookChallenge` in `packages/whatsapp/webhooks/types.ts` still does `token !== verifyToken`.

## Change

Swap that compare for `crypto.timingSafeEqual` after a length check (`timingSafeEqual` throws if lengths differ). Import `crypto` from `node:crypto`. Keep `{ valid: false, statusCode: 403 }` on failure and `{ valid: true, challenge }` on success.

Leave `mode !== 'subscribe'` and the `!verifyToken` guard alone. Do not touch `verifyWhatsappWebhookSignature`. `packages/vapi/webhooks/types.ts` already has this pattern.

## Check

Add two tests in `packages/whatsapp/webhooks.test.ts`: match still valid, mismatch / length mismatch still 403.

thisisanubhav:
## Description

Fixes #704.

`verifyWhatsappWebhookChallenge` in `packages/whatsapp/webhooks/types.ts` compared the incoming `hub.verify_token` against the configured value with `!==`. String comparison short-circuits on the first differing byte, so response timing leaks the token byte-by-byte to anyone who can hit the webhook verification endpoint.

This PR swaps that compare for `crypto.timingSafeEqual` behind an equal-length check (`timingSafeEqual` throws on unequal lengths), matching the pattern already in `packages/vapi/webhooks/types.ts` — which the issue points at — and in `packages/instagram/webhooks/challenge.ts`, the closest analogue since Instagram uses the same Meta `hub.verify_token` handshake:

```ts
if (mode !== 'subscribe' || !verifyToken || !token || !challenge) {
	return { valid: false, statusCode: 403 };
}
// Constant-time compare so a mismatch can't be narrowed byte-by-byte from
// response timing. timingSafeEqual throws on unequal lengths, so check first.
const a = Buffer.from(token);
const b = Buffer.from(verifyToken);
if (a.length !== b.length || !timingSafeEqual(a, b)) {
	return { valid: false, statusCode: 403 };
}
return { valid: true, challenge };
```

**One deviation from the issue text, flagged for review.** The issue says to leave the first guard alone, but I added `!token` to it. `token` is `string | undefined`, and the old `token !== verifyToken` was implicitly covering the undefined case; without `!token`, `Buffer.from(undefined)` throws. Outcomes are unchanged — a missing token was a 403 before and is a 403 now — but the guard has to carry that check explicitly once the comparison moves below it.

Everything else is untouched: the `mode !== 'subscribe'` and `!verifyToken` guards, the `{ valid: false, statusCode: 403 }` failure shape, the `{ valid: true, challenge }` success shape, and `verifyWhatsappWebhookSignature`. Scope is a single plugin package (`packages/whatsapp/`).

## Tests

Six cases added to `packages/whatsapp/webhooks.test.ts` under a new `Challenge Verification` block:

- matching verify token → `{ valid: true, challenge }`
- equal-length token mismatch → 403
- **different-length token mismatch → 403 rather than throwing** — asserted with `.not.toThrow()`, since this is the case that regresses if the length check is ever dropped ahead of `timingSafeEqual`
- missing `hub.verify_token`, and empty configured token → 403
- `hub.mode` other than `subscribe`, and missing `hub.challenge` → 403
- repeated query param → first value is read

As in #855, the outcome-level assertions pass against both the old `!==` and the new `timingSafeEqual` implementation — they lock in that the rewrite preserves every valid/invalid result, while the security property itself comes from `timingSafeEqual`. The length-mismatch test is the one genuine behavioural guard, and it fails against a naive `timingSafeEqual` without the length check.

## Checklist

- [x] I have added or updated necessary documentation (none needed — no public surface or behavioural change)

No UI or CLI surface to record: this is a security hardening change to an internal webhook verification helper, and the third-party handshake it guards can't be driven from CI. Per the precedent set by #855, the working proof is the unit test run on this PR:

https://github.com/corsairdev/corsair/pull/890/checks

which executes `packages/whatsapp/webhooks.test.ts` in the "CI Checks" job. Local run of the same suite:

```
PASS ./webhooks.test.ts
  WhatsApp Webhooks
    Signature Verification
      ✓ should verify correct signature via HMAC SHA256 (1 ms)
      ✓ should reject when verification fails
    Challenge Verification
      ✓ should accept a matching verify token and echo the challenge (1 ms)
      ✓ should reject a same-length token mismatch with 403
      ✓ should reject a length mismatch with 403 rather than throwing
      ✓ should reject a missing verify token with 403
      ✓ should still reject a non-subscribe mode or missing challenge
      ✓ should read the first value when a query param repeats
    Message Handler
      ✓ should extract text from an inbound image message with caption (207 ms)

Tests:       9 passed, 9 total
```

Also run locally: full `packages/whatsapp` suite 12/12 pass, `tsc --build --force` clean, `tsup` build succeeds, `biome check` clean, and `pnpm run validate:plugins` passes.

## Additional Notes

No breaking changes, no new dependencies — `node:crypto` is already used across the webhook verifiers. `verifyWhatsappWebhookChallenge` is re-exported from `packages/whatsapp/index.ts` but has no other in-repo callers, so nothing downstream changes behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WhatsApp webhook challenge validation with more secure token comparisons.
  * Rejected blank or invalid verification tokens and handled repeated parameters consistently.
  * Continued returning appropriate rejection responses for missing or invalid challenge parameters.
  * Added coverage for valid requests, token mismatches, missing values, invalid modes, repeated parameters, and other edge cases.

### DIFF
--- a/packages/whatsapp/webhooks.test.ts
+++ b/packages/whatsapp/webhooks.test.ts
@@ -147,6 +147,21 @@ describe('WhatsApp Webhooks', () => {
 
 			expect(result).toEqual({ valid: true, challenge: '1234567890' });
 		});
+
+		it('should reject when a repeated verify token matches only after the first value', () => {
+			// Parameter-pollution guard: reading anything but index 0 — switching
+			// value() to raw.find()/includes(), say — would let an appended query
+			// param smuggle a valid token past verification. Both entries are the
+			// same length, so this exercises timingSafeEqual, not the length check.
+			const result = verifyWhatsappWebhookChallenge(
+				challengeQuery({
+					'hub.verify_token': ['wrong-token-val', 'my-verify-token'],
+				}),
+				'my-verify-token',
+			);
+
+			expect(result).toEqual({ valid: false, statusCode: 403 });
+		});
 	});
 
 	describe('Message Handler', () => {
--- a/packages/whatsapp/webhooks.test.ts
+++ b/packages/whatsapp/webhooks.test.ts
@@ -1,5 +1,8 @@
 import { verifyHmacSignatureWithPrefix } from 'corsair/http';
-import { verifyWhatsappWebhookSignature } from './webhooks/types';
+import {
+	verifyWhatsappWebhookChallenge,
+	verifyWhatsappWebhookSignature,
+} from './webhooks/types';
 
 jest.mock('corsair/http', () => ({
 	verifyHmacSignatureWithPrefix: jest.fn(),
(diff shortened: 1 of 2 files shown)


ambikeesshh:
@greptileai review

devjain32 (approved):

PR #1057, opened 2026-08-25
fix(cli): report async command failures cleanly instead of an unhandled rejection

### LINKED ISSUE #1056: CLI: async command failures surface as an unhandled promise rejection
### What happens

The CLI entrypoint parses argv with the synchronous `Command.parse()`:

```ts
// packages/cli/src/index.ts
if (detectIsMainModule()) {
	createProgram().parse(normalizeLegacyArgs(process.argv));
}
```

But every command action is wired as `async` in `base.command.ts`:

```ts
command.action(async (...data) => {
	...
	await this.action({ args, options: commandOptions });
});
```

`.parse()` does not await the promise returned by an async action, and there is no top-level handler. So when an action rejects — e.g. a failed OAuth token exchange during `corsair setup`, or a network error in `watch-renew` — Node prints a raw `UnhandledPromiseRejection` stack trace instead of the clean `[#corsair]: ...` diagnostic the rest of the CLI uses (12 call sites, e.g. `utils/corsair-instance.ts`).

### Impact

Confusing UX on every failing command; the process can also exit `0` on older Node versions before the rejection is reported. Affects `setup`, `auth`, `list`, `schema`, `watch-renew`, `studio`, `script`, and the subscribe commands.

### Expected

A rejecting action should print a single `[#corsair]: <message>` line and exit non-zero.

### Fix

Use `.parseAsync()` with a top-level `.catch()`. Commander already handles its own parse/help/version exits synchronously, so this only affects genuine action errors. PR attached.

yashksaini-coder:
Closes #1056.

## Problem

`packages/cli/src/index.ts` parsed argv with the synchronous `.parse()`, but every command action is wired as `async` in `base.command.ts`. `.parse()` never awaits the returned promise and there is no top-level handler, so a rejecting action (failed OAuth token exchange in `setup`, network error in `watch-renew`, …) escapes as a raw `UnhandledPromiseRejection` stack dump instead of the `[#corsair]: ...` diagnostic used everywhere else in the CLI.

## Fix

Switch to `.parseAsync()` and attach a `.catch()` that prints one `[#corsair]:` line and exits `1`. The formatter is a small `utils/format-cli-error.ts` helper (mirrors the existing inline format in `utils/corsair-instance.ts`) so it can be unit-tested without importing the `import.meta` entrypoint.

```ts
createProgram()
	.parseAsync(normalizeLegacyArgs(process.argv))
	.catch((err: unknown) => {
		console.error(formatCliError(err));
		process.exit(1);
	});
```

Commander handles its own parse-error / `--help` / `--version` exits **synchronously** (no `.exitOverride()` anywhere), so those paths are untouched — only genuine action rejections now route through the catch. No behavior change on the success path, no migration.

## Testing

- `pnpm --filter @corsair-dev/cli typecheck` — clean
- `pnpm --filter @corsair-dev/cli test` — 11 passed (3 new for `formatCliError`)
- `biome check` on changed files — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI handling for command parsing failures.
  * Errors now display as consistently formatted diagnostic messages and exit with status code 1.
  * Multiline messages are condensed into a single readable line.
  * Prevented duplicate error prefixes and removed trailing whitespace from diagnostics.

* **Tests**
  * Added coverage for successful commands, parsing failures, varied error values, undefined failures, and multiline messages.
  * Verified that diagnostics include a single expected prefix.


yashksaini-coder:
Addressed the CodeRabbit note in c0dd5459: `formatCliError` now collapses newline runs so a multiline error message stays a single `[#corsair]:` line, with a regression test. Left the docstring-coverage warning as-is — the touched functions are covered and the Plugin PR Gate passes; padding docstrings to hit the 80% bar would just be noise.

ambikeesshh:
@greptileai review

ambikeesshh (approved):
LGTM

devjain32 (approved):

PR #1058, opened 2026-08-25
test(corsair): cover verifyHmacSignature and verifyHmacSignatureWithPrefix

### LINKED ISSUE #1059: test: verifyHmacSignature and verifyHmacSignatureWithPrefix lack unit coverage
### Gap

`packages/corsair/async-core/webhook-utils.ts` exports four verification helpers, but the existing `tests/webhook-utils.test.ts` only covers two of them:

| helper | tested before |
| --- | --- |
| `verifyHmacSha256Signature` | ✅ |
| `verifySlackSignature` | ✅ |
| `verifyHmacSignature` | ❌ |
| `verifyHmacSignatureWithPrefix` | ❌ |

`verifyHmacSignature` and `verifyHmacSignatureWithPrefix` are the generic HMAC primitives that back many plugin webhook verifiers via `corsair/http`, so a regression here would silently weaken signature checks across those plugins.

### What coverage should assert

For both helpers:

- fail closed on an empty secret and on an empty signature
- accept a correctly computed signature
- reject an equal-length but non-matching signature (the `crypto.timingSafeEqual` **returns-false** path)
- reject a length-mismatched signature — the case where `crypto.timingSafeEqual` **throws** and the `catch` must return false
- `verifyHmacSignature`: the `sha1` algorithm variant
- `verifyHmacSignatureWithPrefix`: reject a signature missing the expected prefix; accept a valid prefixed signature

Signatures should be computed in-test (no hardcoded secret literals, to avoid the repo secret scanner).

Fix in #1058.

yashksaini-coder:
Closes #1059.

## What

Adds unit tests for the two shared HMAC helpers in `packages/corsair/async-core/webhook-utils.ts` that had no coverage:

- `verifyHmacSignature`
- `verifyHmacSignatureWithPrefix`

The existing `tests/webhook-utils.test.ts` only covered `verifyHmacSha256Signature` and `verifySlackSignature`, yet these two primitives back many plugin webhook verifiers through `corsair/http` — so they're worth pinning down.

## Coverage

For each helper:

- fails closed on an empty secret and on an empty signature
- accepts a correctly computed signature
- rejects an equal-length but non-matching signature (the `timingSafeEqual` **returns** false path)
- rejects a length-mismatched signature — the case where `crypto.timingSafeEqual` **throws** and the `catch` returns false
- `verifyHmacSignature`: sha1 algorithm variant
- `verifyHmacSignatureWithPrefix`: rejects a signature missing the expected prefix; accepts a valid prefixed signature

Signatures are computed in-test with `crypto.createHmac`; no hardcoded secret literals (avoids the repo secret scanner).

## Testing

- `pnpm --filter corsair test -- tests/webhook-utils.test.ts` — 15 passed
- `biome check` — clean

No production code changed.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for HMAC signature verification.
  * Added tests for SHA-1 and SHA-256 algorithm handling.
  * Added validation for empty, mismatched, malformed, and incorrectly prefixed signatures.
  * Confirmed verification safely rejects invalid signatures without errors.


ambikeesshh:
@greptileai review

ambikeesshh (commented):
LGTM!

devjain32 (approved):

PR #1227, opened 2026-08-27
ci: scope PR Checks on main from the push diff

ambikeesshh:
## Summary
- Plugin PRs already only build/test that plugin + corsair. Pushes to `main` still always ran the full graph (~30 min per merge).
- Classify `main` with the same rules, using `git diff` of the push `before`/`after`. One-plugin lands should log `CI lane: plugin`.
- If the diff is unknown (unborn SHA, missing range, git failure), fail closed to full **including www**. A known empty diff stays packages-only full.

## Test plan
- [ ] Classifier tests: `node --experimental-strip-types --test scripts/pr-review/pr-scope.test.ts`
- [ ] After merge, a one-plugin land on `main` logs `CI lane: plugin` and finishes in ~1–2 min, not ~30
- [ ] A core or two-plugin land on `main` still logs `CI lane: full`
- [ ] This PR itself is a workflow/scripts change, so its own CI is the full lane

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI checks for push-triggered changes by accurately identifying affected files and selecting the appropriate review scope.
  * Added safer handling for incomplete, invalid, or initial push events, with a comprehensive fallback when the change range cannot be determined.
  * Improved cache handling so it is used only for full-scope checks, helping ensure reliable validation results.

* **Tests**
  * Added coverage for push-event processing, revision detection, changed-file resolution, and scope classification.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -43,6 +43,8 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
 
       - uses: pnpm/action-setup@v4
         with:
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -74,7 +74,7 @@ jobs:
         run: pnpm install --frozen-lockfile --filter . --filter "@corsair/www..."
 
       - name: Turbo cache for main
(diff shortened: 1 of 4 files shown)


ambikeesshh (commented):
greptile + coderabbit both want constants.ts to kick a full rebuild. that’s how we got the 30 min runs in the first place. ignoring em

devjain32 (approved):

PR #1449, opened 2026-09-01
fix(corsair): pass a test kek after resolveKekAtInit

Dhirenderchoudhary:
## Summary
-  throws when `kek` is empty and a database is set. `hooks` and `retry-recover` still passed `kek: ''`, so every plugin PR that touches `constants.ts` fails `corsair#test`.
- Use `kek: 'test-kek'` like the other core tests.

## Test plan
- [ ] `corsair#test` (`hooks.test.ts`, `retry-recover.test.ts`)
- [ ] Re-run CI on an open plugin PR after this lands

## Summary by CodeRabbit

* **Tests**
  * Updated test configurations to use a representative encryption key value.
  * Existing test behavior and assertions remain unchanged.

### DIFF
--- a/packages/corsair/tests/hooks.test.ts
+++ b/packages/corsair/tests/hooks.test.ts
@@ -61,7 +61,7 @@ describe('Endpoint Hooks', () => {
 			const modifiedText = 'Modified by before hook';
 
 			const corsair = createCorsair({
-				kek: '',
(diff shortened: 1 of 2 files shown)


Dhirenderchoudhary:
Closing as #1450  same fix done by @yuvrxj-afk  with better approach

PR #1491, opened 2026-09-04
feat(mastra): add @corsair-dev/mastra ToolProvider for Mastra

yuvrxj-afk:
## What

Adds **`@corsair-dev/mastra`** — a standalone [Mastra](https://mastra.ai) `ToolProvider` that exposes Corsair's 200+ integrations to Mastra agents, backed by managed OAuth and with credentials that stay in your own database.

Each Corsair plugin becomes a toolkit and each operation a tool. It extends `BaseToolProvider`, so one provider drives both consumption paths:

- **Editor / managed** — register with `MastraEditor`; the VNext surface (`resolveToolsVNext` / `listToolsVNext` / `listToolkitsVNext`) drives the managed-connection UI (same slot Composio/Arcade fill).
- **Code-config** — call `resolveTools()` and hand the tools to an `Agent` (via the base-class shim).

## Tenancy

`tenantId` is a single knob that maps a Mastra request to a Corsair tenant **and** derives the Mastra connection scope — no separate scope wiring:

| `tenantId`              | Meaning                                 | Derived scope       |
| ----------------------- | --------------------------------------- | ------------------- |
| `'acme'` (string)       | pin every request to one tenant         | `'shared'`          |
| `(req) => …` (function) | resolve per request (multi-tenant SaaS) | `'caller-supplied'` |
| _omitted_               | connection → author id → `'default'`    | `'per-author'`      |

## Package shape

Thin and standalone — a sibling of `@corsair-dev/mcp`, **not** stacked on it. Depends only on `corsair` + `zod`, with `@mastra/core` as the single peer. The provider and its `form-field-to-zod` helper live here; nothing imports `@corsair-dev/mcp`, so bundlers of `@corsair-dev/mastra` don't drag in the other framework adapters.

## Verified

- `tsc --noEmit` against `@mastra/core@1.64` `BaseToolProvider` (all abstract methods + types)
- biome format + lint clean
- pure-function check + real-instance check (`corsair.manage.connectionStatus`, not a stub)
- runtime against a real corsair + plugin: `resolveTools` / `resolveToolsVNext` / `listToolsVNext` / `listToolkitsVNext` / `getHealth` / allowlist filtering build 52 real GitHub tools; derived scope confirmed

## Follow-up

A live `MastraEditor` E2E (real managed OAuth through the editor UI) is the last verification before pitching Mastra to add this to their integrations list — same way the code-config path was proven end to end against real Slack channels.

## Summary by CodeRabbit

- **New Features**
  - Added the `@corsair-dev/mastra` adapter, enabling Mastra agents to use Corsair integrations, dynamic tools, managed OAuth, tenant-aware connections, and connection management.
  - Added a Slack-focused Mastra demo with dynamic tool access.

- **Documentation**
  - Added setup and usage guides for the Mastra adapter and demo.

- **Chores**
  - Expanded workspace, build, test, deployment, and publishing workflows to include adapter packages.
  - Configured the demo to use the Corsair LLM endpoint.


yuvrxj-afk:
@greptileai

devjain32 (approved):

yuvrxj-afk:
@greptileai please re-review.

Addressed since the last pass:
- **Typing rule (R7):** `resolveToolsVNext` returns `ToolAction<any, any, any>` because that is the exact signature of the `@mastra/core` abstract base method it overrides (tool input/output schemas are derived per-plugin at runtime, so they can't be statically narrowed). Documented this on the method per the rule.
- **Adapter checks in CI:** the package `test` script (the `.check.ts` files) now runs via turbo's `test` task in the full lane of `pr-checks.yml`.
- Merged latest main.

Note: `@mastra/core` is intentionally an *optional* peer — the `mcp` package re-exports this adapter via a `./mastra` subpath, so a required peer would force `@mastra/core` on every mcp consumer that doesn't use Mastra.

yuvrxj-afk:
@greptileai

yuvrxj-afk:
@greptileai re-review please — both remaining findings are resolved.

**The root cause (the type regression you flagged):** `AnyCorsairInstance` was instantiating the client generics with a fake shallow plugin shape, which expanded the per-plugin mapped types to a structure no real `createCorsair()` result matches — so real instances were rejected (TS2322) and callers had to cast. Fixed in `packages/corsair/inspect.ts` by parameterizing with an **empty plugin set** (`CorsairSingleTenantClient<readonly []> | …`): that's the shape common to every instance, so any concrete client is a structural superset and assigns cleanly. Real `createCorsair()` values now satisfy the public `corsair` config field with **no cast**.

**Casts:** the widening `as AnyCorsairInstance` casts in the live check are removed (no longer needed); the one remaining `as unknown as AnyCorsairInstance` is on a hand-built test double and now carries a rationale comment per CONTRIBUTING §222–225. The internal `manage`/`withTenant` casts were already documented.

Verified: core `tsc` 0, mastra `tsc` 0, `@corsair-dev/mastra` checks pass, frozen-install exit 0.

yuvrxj-afk:
@greptileai

PR #1503, opened 2026-09-05
feat(sdk): add manage.disconnect

yuvrxj-afk:
corsair.manage.disconnect({ plugin, tenantId }) deletes the account
(encrypted credentials) and its account-scoped entities and events in
one transaction. Idempotent; the approval queue is left to its own
(tenant, plugin) lifecycle. Exposed over HTTP as POST /disconnect.

## Summary by CodeRabbit

* **New Features**
  * Added connection disconnection through the management API and `corsair.manage`.
  * Disconnecting removes stored connection data and related account activity for the selected plugin and tenant.
  * Repeated disconnection requests complete safely when no connection exists.
  * Unknown plugins are rejected with an appropriate error.

* **Bug Fixes**
  * Improved handling of empty or invalid request bodies with clear validation errors.
  * Ensured disconnection remains isolated to the selected tenant and plugin.
  * Catalog integration lookups now return empty results when retrieval fails.

* **Tests**
  * Added coverage for cascading removal, duplicate accounts, idempotency, tenant isolation, and validation.


yuvrxj-afk:
@greptileai

devjain32 (approved):

yuvrxj-afk:
@greptileai please re-review.

Addressed since the last pass:
- **Disconnect race (the 3/5 blocker):** the delete now matches accounts by predicate *inside* the transaction (tenant_id + integration_id), instead of a pre-snapshotted id list — so a row a racing connect commits before the delete is cleared too, and a successful response can't leave live creds behind.
- **Input validation:** a non-string `plugin`/`tenantId` now returns a 400 instead of throwing a TypeError on `.trim()` (added a test).
- Merged latest main; added a build-time guard so the catalog-backed sitemap builds without DATABASE_URL (fixes the Vercel preview).

yuvrxj-afk:
@greptileai Here's how each of the 3 open findings was resolved (commit 7ac71fe8):

**Finding 1 — catalog-integration-cache.ts fallbacks mask runtime DB failures**

Added `const noDB = !process.env.DATABASE_URL` at module scope. Each `try/catch` now returns the fallback only when `noDB` is true (i.e. no DB is configured at build time, exact parity with Vercel preview). When a DB _is_ configured, the error is re-thrown so runtime outages propagate. `getCatalogIntegrationById`'s pre-existing catch is gated the same way for consistency. Build without `DATABASE_URL` still exits 0 and prerenders `/sitemap.xml`.

**Finding 2 — disconnectConnection concurrent-connect race (STOPPING, reporting options)**

After tracing the full call graph, I'm not applying a fix here. Here's why and what the options are:

The two account-creation paths (`ensureAccount` in `oauth/index.ts:81` and `ensureIntegrationAccountRow` in `hub/internal/account.ts:5`) both do check-then-insert without a unique constraint on `(tenant_id, integration_id)`. The existing delete-by-predicate inside a transaction clears rows committed _before_ the delete. The forward-racing case (insert commits _after_ the delete) is real but the correct fix has two viable shapes:

- **Option A — DB unique constraint**: Add `UNIQUE(tenant_id, integration_id)` to `corsair_accounts` (migration). `ensureAccount` already calls `.onConflict(oc => oc.column('id').doNothing())` on the PK; a `(tenant_id, integration_id)` unique constraint would need both writers updated to `ON CONFLICT DO NOTHING/UPDATE`. Portable to SQLite and Postgres. This is the clean root-cause fix but touches a migration + two writers.
- **Option B — Postgres advisory lock**: `pg_try_advisory_xact_lock(hashtext(tenant_id || integration_id))` at the top of both the disconnect transaction and the account-creation transaction serializes the lifecycle per pair. Zero-migration, but advisory locks are Postgres-only and the test DB is SQLite (`better-sqlite3`), so tests would need a mock/skip — introducing fragility.

Recommending Option A. Not applying it here because it requires a schema migration and coordinated changes to two writers that are out of scope for this PR.

**Finding 3 — `as unknown as string` typing-rule violation in disconnect.test.ts**

Replaced the inline field-level double-cast (`plugin: 1 as unknown as string`) with a single boundary cast on the whole input object: `const badInput = { plugin: 1 } as unknown as Parameters<typeof disconnectConnection>[1]`. This matches the established convention across the test suite (cast the fixture at the argument/object boundary, not individual fields). The assertion still proves the 400/bad_request behavior, not a TypeError.

yuvrxj-afk:
@greptileai please re-review.

Resolved:
- **Catalog fallback masking runtime failures:** the empty fallback now only fires when no database is configured at build time (`!process.env.DATABASE_URL`); when a DB is configured, the error propagates so real runtime outages surface.
- **Test typing:** replaced the field-level `as unknown as string` with a single boundary cast on the whole input, matching the suite's convention.

**Deferred (deliberately, tracked):** the concurrent-connect race. It's a pre-existing account-lifecycle issue — `ensureAccount` / `ensureIntegrationAccountRow` do check-then-insert with no `(tenant_id, integration_id)` unique constraint, so a connect committing *after* the delete can leave a live row. The predicate-in-txn delete here already closes the commit-before-delete window; fully closing the after-delete window needs a unique constraint + upsert on those writers, which is out of scope for this PR and tracked as a follow-up. Documented inline in `disconnectConnection`.

yuvrxj-afk:
@greptileai please re-review — the concurrent-connect race you flagged is now fully closed (SDK-only, no schema change).

**What changed since the last review:**
- `disconnectConnection`, `ensureAccount` (oauth/index.ts) and `ensureIntegrationAccountRow` (hub/internal/account.ts) now all take a `SELECT … FOR UPDATE` lock on the shared `corsair_integrations` row *before* their read-then-write, so a connect and a disconnect for the same plugin can no longer interleave and leave a live credential behind a `disconnected: true`.
- The lock is guarded by `isPg` on `CorsairDatabase` (set from the existing dialect detection): Postgres emits `FOR UPDATE`; SQLite/pglite skip it — its single-writer model already serializes writes, so no invalid SQL.
- No unique constraint / migration — core does not own the user's DB schema; serialization is the portable root-cause fix.
- Removed the stale "Known limit" comment; the writers build the integration select once and conditionally chain `.forUpdate()` (no duplicated query).

Verified: `tsc --noEmit` exit 0; `disconnect` + `management-handler` 22/22; connect/oauth account-creation paths 55/55.

yuvrxj-afk:
@greptileai the last finding is resolved — the two `as unknown as` casts in `disconnect.test.ts` now carry the narrow, commented rationale that CONTRIBUTING.md requires (why each is safe + why a better type wasn't practical). Nothing else changed. Should be clean to 5/5 now.

yuvrxj-afk:
@greptileai the two remaining `as any` casts in `management-handler.test.ts` (the JSON `config` column on the seed inserts) are now documented in the file-level cast-rationale block, per CONTRIBUTING. All `unknown`/`any` assertions in the PR now carry their rationale.

PR #1725, opened 2026-09-10
feat(core): tool-call route (/call) behind an auth gate

yuvrxj-afk:
Exposes the SDK's tool operations over HTTP so any language can drive Corsair through the management handler.

## What
- `POST /:tenant/:plugin/call/:op` invokes an operation; `GET /call` lists the registered plugin→op set.
- Both are **off by default**. They turn on when the host owns access control: either `unsafeAllowUnauthenticated: true` (the developer's own backend fronts the handler) or by supplying an `authenticate` hook (e.g. a Bearer project key), which also enables the route.
- `authenticate` fronts every route except the ones flagged `public`: health, and the OAuth-callback / connect-resolve / permission-lookup legs that arrive without the project key and are guarded by an unguessable state/token instead. Hub token delivery is verified by signature at the base path and is unaffected.

## Core
- `resolveCall` + `listRegisteredOps`: per-(instance, tenant) client cache so concurrent same-tenant calls share one refresh single-flight; operation-tree walk; error normalization to 401/403/502/500 with response hygiene (never spreads `ApiError.request`, so the Authorization header can't leak to a caller).
- `PermissionRequiredError` gains a `reason` discriminator (`denied` / `policy` / `timeout` / `pending`) so a non-TS caller can distinguish a hard deny from an approval a human still has to grant.

## Tests
22 tests (resolution, error normalization, the auth gate, public-leg bypass). `pnpm check-types` clean for the changed files; full suite green.

Motivation: foundation for the hosted multi-language runtime (SDK served as a REST runtime).

## Summary by CodeRabbit

* **New Features**
  * Added management API routes for listing registered operations and invoking plugin operations.
  * Added optional authentication and unauthenticated-access controls for operation calls.
  * Added tenant-aware operation resolution for single- and multi-tenant setups.
  * Permission errors now include machine-readable reasons: denied, policy, timeout, or pending.

* **Bug Fixes**
  * Improved tenant client retention during concurrent operations and refreshes.
  * Standardized call error handling while preventing sensitive authorization details from appearing in provider errors.

* **Tests**
  * Added coverage for operation calls, authentication, tenant handling, error responses, caching, and permission reasons.


devjain32 (approved):

yuvrxj-afk:
@greptileai

yuvrxj-afk:
@greptileai

PR #1729, opened 2026-09-10
fix(tunnel): supervise frpc and restart on death

yuvrxj-afk:
## What
Supervise the dev frpc tunnel so it restarts on death, and bump `corsair` to `0.1.130`.

## Why
`maybeStartTunnel` spawned frpc once at boot and, on death, only did `activeTunnels.delete(key)` — nothing respawned it. So when frpc died (laptop sleep, network blip, frps restart) the tunnel stayed **down until the app process restarted**, while the Hub kept delivering OAuth tokens / run / probe to the dead URL. That surfaced as "the redirect isn't working" / "the API call times out" even though the app was up.

## Changes
- Extract `superviseTunnel({ start, schedule, minDelayMs, maxDelayMs })`: restart on death with **capped exponential backoff** (1s → 30s); a healthy start resets the backoff. The Hub-owned slug is sticky, so a restart re-registers the **same public URL** and the Hub's stored `tunnel_url` self-heals. Extracted from the spawn so the restart wiring is unit-testable.
- `maybeStartTunnel` wires `runTunnel`'s `onClose` into it; drops the old "give up + delete key" path. Failure now logs `tunnel down: … Retrying — run \`corsair setup\` if it doesn't recover.` and backs off.
- **Dedupe double death-signal:** `runTunnel`'s `fail()` both rejects the promise *and* kills the child, whose `exit` then fires `onClose` — two signals for one dead attempt. `restart` is idempotent per generation so a death can't spawn two overlapping frpc processes.
- `corsair` `0.1.129 → 0.1.130` — publishes this supervisor **and** the delivery-request timeout already on `main` but unpublished under `0.1.129`.

## Testing
- New `tests/supervise-tunnel.test.ts` (4): restart-on-death, double-signal dedupe, capped exponential backoff `[1,2,4,4,4]`, backoff-reset-after-healthy.
- `tsc --noEmit` clean; `biome check` clean; existing `tunnel.test.ts` + `run-tunnel.test.ts` 22/22 unchanged.

## Notes
- Retries indefinitely (30s cap) rather than classifying transient vs permanent errors — reverting to give-up-on-first-failure is the boot-time-blip bug this fixes. The cap bounds the noise and the log is actionable.

## Summary by CodeRabbit

- **Improvements**
  - Development tunnels now automatically restart after unexpected interruptions.
  - Restart attempts use an increasing delay, capped at 30 seconds, to improve stability during repeated failures.
  - Duplicate or delayed failure signals no longer trigger overlapping restarts.
  - Recovery after a healthy tunnel start resets the restart delay to its minimum.

### DIFF
--- a/packages/corsair/core/index.ts
+++ b/packages/corsair/core/index.ts
@@ -208,18 +208,20 @@ export function superviseTunnel(opts: {
 	const min = opts.minDelayMs ?? TUNNEL_RESTART_MIN_MS;
 	const max = opts.maxDelayMs ?? TUNNEL_RESTART_MAX_MS;
 	let delay = min;
-	let scheduled = false;
-	// A dead attempt can signal twice — the promise rejects AND onClose fires
-	// (runTunnel's fail() kills the child, whose exit then runs onClose). Collapse
-	// both into one restart, else each death spawns two overlapping frpc processes.
-	const restart = () => {
-		if (scheduled) return;
-		scheduled = true;
-		schedule(run, delay);
-		delay = Math.min(delay * 2, max);
-	};
-	const run = () => {
-		scheduled = false;
+	const run = (): void => {
+		// One restart per attempt, latched per attempt (not globally). A dead
+		// attempt can signal twice — runTunnel's fail() rejects the promise and
+		// kills the child, whose exit later fires onClose, possibly *after* the
+		// next attempt already started. onClose is bound to this attempt's
+		// `restart`, so a superseded attempt's late signal no-ops here instead of
+		// spawning an overlapping frpc process.
+		let ended = false;
+		const restart = (): void => {
+			if (ended) return;
+			ended = true;
+			schedule(run, delay);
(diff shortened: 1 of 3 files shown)


yuvrxj-afk:
Addressed the delayed-exit race flagged by @greptile-apps and @coderabbitai (both point at the same thing) in `b940f958`.

The prior dedup used a single module-scoped `scheduled` flag, which only collapsed reject+onClose when they fired within one `run()`. A superseded attempt's killed child that exits *after* the next attempt starts would see `scheduled=false` and schedule an extra overlapping frpc.

Fix: latch the one-restart guard **per attempt** (closure-scoped `ended`). `onClose` is bound to its own attempt's `restart`, so a stale attempt's late signal no-ops. Added a `ignores a delayed close from a superseded attempt` regression test (5/5 supervise tests green; tsc + biome clean).

yuvrxj-afk:
@greptileai

devjain32 (approved):

PR #1733, opened 2026-09-14
▚▚ fix(corsair): honor rateLimitConfig.enabled=false and stop retrying 429

breken-ai:
`request()` in `packages/corsair/async-core` retries 429 responses even when the caller passes `rateLimitConfig: { enabled: false }`. A plugin that turns transport retries off (to let its own error handlers own 429, as several in-tree plugins do) still gets `maxRetries` silent retries with exponential back-off — about 7s of hidden delay and three extra calls per rate-limited request under the default `maxRetries: 3` — before the `ApiError` reaches it.

## Root cause
`enabled` is only honoured on the response path: `isRateLimitError()` returns false when disabled, so the 429 falls through to `catchErrorCodes`, which throws an `ApiError`. The catch branch (`request.ts` ~L407 and the fallthrough ~L430) then retries any `ApiError` whose status is 429 without checking `rateLimitConfig.enabled`. The disabled config is in fact the only way to reach that catch-branch retry, so the flag was inverted in practice. Every plugin that sets `enabled: false` today also sets `maxRetries: 0`, which is why the defect stayed hidden.

## Fix
- `packages/corsair/async-core/request.ts`: compute `retryableRateLimit = rateLimitConfig.enabled && error instanceof ApiError && error.isRateLimitError()` once in the catch block and gate both retry sites on it. Behaviour with `enabled: true` is unchanged.

## Tests
- `packages/corsair/tests/request-rate-limit-disabled.test.ts` (new): mocks `fetch` to always answer 429.
  - `enabled: false, maxRetries: 2` must reject with the 429 `ApiError` after exactly one fetch — fails on main with "Received number of calls: 3".
  - `enabled: true, maxRetries: 2` still retries to `maxRetries + 1` fetches (control; passes before and after).
- Full `packages/corsair` jest suite: 82 suites / 642 tests pass; package typecheck and biome clean.

## Not changed
- Plugins that pair `enabled: false` with `maxRetries: 0` keep working as before; they can drop the redundant `maxRetries: 0` later but this PR does not touch them.

---
▚▚ **Shipped by breken** — self-healing software. This one's on us. [breken.ai](https://breken.ai)

## Summary by CodeRabbit

- **Bug Fixes**
  - Rate-limit handling now respects the rate-limit setting. When disabled, rate-limit errors are returned immediately without retries.
  - When enabled, rate-limit errors continue to retry according to the configured limit.

- **Tests**
  - Added coverage verifying both disabled and enabled rate-limit retry behavior.


packages/corsair/async-core/request.ts
```diff
@@ -427,7 +434,7 @@ export const request = <T>(
 					return;
 				}
 
-				if (error instanceof ApiError && error.isRateLimitError()) {
+				if (retryableRateLimit) {
```
ambikeesshh, line 437: this second branch looks dead to me. first block already continues when we have attempts left, and the maxattempts check above already rejects otherwise. leaving it is fine for now, lets just clean it up separately so the next person doesnt think there are two retry paths here.

packages/corsair/tests/request-rate-limit-disabled.test.ts
```diff
@@ -0,0 +1,69 @@
...
+		const fetchMock = mockAlways429();
+
+		const promise = request(config, options, {
+			rateLimitConfig: { ...baseRateLimitConfig, enabled: false },
+		});
+
+		await expect(promise).rejects.toMatchObject({ status: 429 });
+		await expect(promise).rejects.toBeInstanceOf(ApiError);
+		expect(fetchMock).toHaveBeenCalledTimes(1);
+	});
```
ambikeesshh, line 57: can we grab the error once with try/catch and assert both props there, instead of awaiting the same rejected promise twice. works as is, just matches how we usually write these here.

PR #1739, opened 2026-09-15
feat(cloud): createCorsairCloud — hosted-runtime client + contract + stale-token fix

yuvrxj-afk:
Ships the **client half of Corsair Cloud** and the frozen HTTP contract every language wrapper will build on. (The multi-language client *codegen* was pulled out in `61e94dbf` — see "Deferred" — but the **contract stays**.)

## What it ships

**`createCorsairCloud({ apiKey, url, signingSecret? })`** — a dedicated cloud surface, separate from `createCorsair` (which is untouched):
- **No plugin list, no kek/db** — the plugin set lives on the VM. Multi-tenant **by default**.
- **Dynamic client**: `corsair.withTenant(t).<plugin>.api.<op>(args)` → `POST /{tenant}/{plugin}/call/{op}` on the hosted runtime. `manage.*` covers tenants / connect / connectionStatus / disconnect.
- **Augmentable typing**: `CorsairCloudRegistry` — `any` until populated, then fully typed via an inline generic now or a CLI-generated `.d.ts` later. Autocomplete without importing plugins.
- Internally, `buildCloudClient`'s plugin list is optional and `buildCloudSurface` is shared with the `ck_cloud_` auto-detect path in `createCorsair`.

**Frozen HTTP contract** — `core/cloud/contract.openapi.yaml` (full runtime surface) + `contract.test.ts` that asserts it stays in sync with `CLOUD_ROUTES`. This is the single source of truth for every future language/framework wrapper.

**Stale-token fix** — the account key-manager cached the account row for the whole process lifetime, so a long-lived `/call` client served a revoked token after an out-of-band token (re)delivery → `AuthMissingError` → false "needs reconnect". Now re-reads the account per credential access (DEK cached, keyed to its ciphertext). Regression test added. (Only manifested in cloud: local dev/prod has no long-lived multi-tenant runtime + `/call` client cache + out-of-band delivery.)

## How you use it
```ts
const corsair = createCorsairCloud({ apiKey: 'ck_cloud_…', url: 'https://<vm>.corsair.cloud/<env>/api/corsair' });
await corsair.withTenant('acme').notion.api.pages.searchPage({});
await corsair.manage.connect.createLink({ plugin: 'notion', tenantId: 'acme' });
```

## Deferred (next: the wrappers track)
- Multi-language clients (Swift / Python / Go / Rust) — built **on `contract.openapi.yaml`**, as thin idiomatic wrappers, not raw generated clients.
- `CorsairProvider` + `useCorsair() → { api, db }` for React (client-side; secret injected by a backend proxy, never in the browser).
- `databaseUrl` on the cloud config (BYO / hosted Postgres), consumed by the runtime.

## Verification
74 tests green; src typecheck clean; local adversarial review done (dropped a redundant `/call`-cache eviction that could split the refresh single-flight).


yuvrxj-afk:
@greptileai

yuvrxj-afk:
Most inline findings were on `2d032c8f`; status against latest (`d846f351`):

**Greptile** @greptileai
- KEK guard breaks local clients — **fixed**: removed the construction-time throw; enforcement moved to `signState`/`verifyAndDecodeState`, so an empty KEK can never sign/verify a forgeable OAuth state while construction stays lenient.
- Plugin schema differs from runtime — **fixed**: `PluginInfo` corrected to `{id, authType, configured, missingFields, oauth}`.
- ReDoS in `http.ts` — **fixed**: regex-free slash trim.
- Cloud discovery not exposed — **intentional/deferred**: removed the unreachable code from the typed TS surface; the runtime and generated clients still expose it. Follow-up: surface via the base `manage` in both in-process and cloud modes.

**CodeRabbit** - @coderabbitai
- KEK omission contract — **fixed** (see above).
- `.db` in docs surface — **moot**: design docs removed from the repo (kept local).
- Empty/malformed 2xx JSON → `undefined` cast — **valid, addressing**: will throw a typed `internal_error`.
- `createLink({})` can violate the contract — **valid, addressing**: cloud input will require `plugin`+`tenantId`.
- Request timeout / `AbortSignal` — **valid, addressing**.
- Mock-runtime listen race — **valid, addressing**: await the `listening` event.
- Pin openapi-generator version — **valid, follow-up**.
- `CORSAIR_CLOUD_URL` test env restore / docstring coverage — noted (minor).

yuvrxj-afk:
Both P1s fixed in `a09b7450`:
- **Cleartext key** — `resolveCloudBaseUrl` now requires `https://`; `http://` is allowed only for loopback (localhost/127.0.0.1/::1) so the mock-runtime tests still run. Non-loopback `http` throws.
- **Multi-tenant connectionStatus** — `connectionStatus.get()` now throws without a tenantId in multi-tenant mode; the `default` fallback is kept for single-tenant only.

Cloud + kek tests 49/49, `tsc --noEmit` clean.

devjain32 (approved):

yuvrxj-afk:
Pushed two commits:

- **feat(cloud): `createCorsairCloud`** — dedicated cloud surface: no plugin list, multi-tenant by default, dynamic client (`corsair.withTenant(t).<plugin>.api.<op>(args)` → HTTP). Types come from an augmentable `CorsairCloudRegistry` (inline generic now; CLI-generated `.d.ts` later) so autocomplete works without importing plugins. `buildCloudClient` plugin list is now optional; `buildCloudSurface` shared with the existing `ck_cloud_` auto-detect path.
- **fix(cloud): stale-token false reconnect** — the account key-manager cached the account row for the whole process lifetime, so a long-lived `/call` client served a revoked token after an out-of-band token (re)delivery → `AuthMissingError` → false "needs reconnect". Now re-reads the row per credential access (DEK cached, keyed to its ciphertext). Regression test added.

Local adversarial review done: dropped a redundant cache-eviction layer (could split the refresh single-flight), trimmed comments. 74/74 tests green, src typecheck clean.

@greptileai please re-review the two new commits — focused on the key-manager re-read (any torn-read in `doUpdateConfig`) and the `createCorsairCloud` type surface.

yuvrxj-afk:
Both findings addressed in 417805ce:
- **P1 DEK-race discards credentials** — `doUpdateConfig` now reads the account row **once** and derives the DEK + current config from that same snapshot, then re-encrypts with that DEK. A concurrent DEK rotation can no longer leave the DEK and config mismatched (which was what tripped the decrypt-failure catch that reset config to {}). The stale-token re-read fix is what surfaced the two-read window; this closes it.
- **Loose types documented** — the dynamic op proxy in client.ts now carries the rationale (op set + arg/return shapes are VM-defined, discovered at runtime; the typed layer is CorsairCloudRegistry). The `as unknown as` casts already note the cloud-covers-a-subset reason.
47 tests + src typecheck green.

PR #1742, opened 2026-09-16
fix(cloud): gate cloud-client mode on a base URL, not the ck_cloud_ key prefix (0.1.136)

yuvrxj-afk:
## Problem (prod regression in 0.1.135)
Hosted Corsair Cloud VMs on `corsair@0.1.135` crash-loop at startup and every project apply fails with "Configuration apply failed on the runtime."

Root cause: `0.1.135` (#1739) made `createCorsair()` switch into cloud-**client** mode whenever the project key is `ck_cloud_`-prefixed. But the hosted **runtime** calls `createCorsair()` with a `ck_cloud_` key while being a full server (plugins + db + kek, no cloud URL). It wrongly enters client mode, demands `CORSAIR_CLOUD_URL`, throws `Cloud mode (ck_cloud_ key) requires a base URL`, and systemd restarts it forever. The apply health-check never gets a 200. Proven `0.1.135`-specific: `0.1.134` runs the same `ck_cloud_` keys fine.

## Fix
- **`core/index.ts`**: enter cloud-client mode only when the key is `ck_cloud_` **and** a cloud base URL resolves (`hub.baseUrl` or `CORSAIR_CLOUD_URL`). No URL → build the normal server client (the runtime's case). A real cloud client always has a URL, so this doesn't affect client usage.
- **`hub/config.ts`**: second-order fix — the fall-through path still unconditionally required `signingSecret`, which the runtime never sets (would crash-loop on a different error). Exempt `ck_cloud_` keys, restoring `0.1.134` behavior.
- `createCorsairCloud` unchanged (explicit cloud client).

## Tests
TDD: runtime-shaped config (`ck_cloud_` key + plugins + db + kek, no baseUrl/signingSecret) now builds a server instead of throwing (watched fail→pass). ck_cloud_ + baseUrl still → cloud client. `ck_dev_`/`ck_prod_` paths confirmed untouched. Cloud suite 45/45; dev/prod suites 60/60; tsc clean.

Ships as `0.1.136` (0.1.135 is published/immutable) — carries createCorsairCloud + the stale-token fix + this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cloud credentials when no cloud base URL is configured, allowing the client to use the standard connection path instead of attempting unavailable cloud routing.
  * Cloud credentials no longer require a signing secret, while standard credentials continue to require one.
  * Improved detection of valid cloud endpoints from configured or environment-provided URLs.

* **Chores**
  * Updated the Corsair package version to 0.1.136.


ambikeesshh (approved):

yuvrxj-afk:
@greptileai

PR #1751, opened 2026-09-19
feat: add Starton integration (gate fix)

yuvrxj-afk:
Supersedes #1598 — restores R1 scope after fork cannot accept workflow file pushes without workflow PAT scope.


yuvrxj-afk:
Opened by mistake during gate-fix attempt on #1598; closing.

PR #1755, opened 2026-09-20
feat(clicksend): add clicksend integration plugin

yuvrxj-afk:
Gate-fix branch with restored workflow scope. See #1700.


yuvrxj-afk:
Duplicate — gate fix belongs on #1700.

